PR #4, opened 2021-12-05
Add ScuttleButt UDP server

### LINKED ISSUE #2: Implement a UDP server to server scuttlebutt

chrisduerr:
This patch adds a new `server` module, providing functionality for two
ScuttleButt instances on different machines to stay in sync.

The new `ScuttleServer` struct provides a UDP server which runs in the
background once spawned with `ScuttleServer::spawn`, automatically
replying to incoming gossip using the shared `ScuttleButt` mutex.

To propagate gossip to other servers, the `ScuttleServer::gossip` method
allows transmitting the current state to other instances. While it would
be possible to keep track of all incoming and outgoing conversations, to
allow for broadcasting updates to all known nodes, this is currently
still left to the consumer.

Fixes #2.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,3 +1,5 @@
+pub mod server;
+
 mod model;
 
 use crate::model::{ClusterState, NodeState, ScuttleButtMessage};
@@ -11,12 +13,11 @@ pub struct ScuttleButt {
 }
 
 impl ScuttleButt {
-
     pub fn with_node_id(self_node_id: String) -> Self {
         ScuttleButt {
             max_transmitted_key_values: 10,
             self_node_id,
-            cluster_state_map: ClusterState::default()
+            cluster_state_map: ClusterState::default(),
         }
     }
 
@@ -24,7 +25,7 @@ impl ScuttleButt {
         self.max_transmitted_key_values = max_transmitted_key_values;
     }
 
-    pub fn create_syn_message(&mut self) -> ScuttleButtMessage {
+    pub fn create_syn_message(&self) -> ScuttleButtMessage {
         let digest = self.cluster_state_map.compute_digest();
         ScuttleButtMessage::Syn { digest }
     }
@@ -32,13 +33,17 @@ impl ScuttleButt {
     pub fn process_message(&mut self, msg: ScuttleButtMessage) -> Option<ScuttleButtMessage> {
         match msg {
             ScuttleButtMessage::Syn { digest } => {
-                let delta = self.cluster_state_map.compute_delta(&digest, self.max_transmitted_key_values);
+                let delta = self
+                    .cluster_state_map
+                    .compute_delta(&digest, self.max_transmitted_key_values);
                 let digest = self.cluster_state_map.compute_digest();
                 Some(ScuttleButtMessage::SynAck { delta, digest })
             }
             ScuttleButtMessage::SynAck { digest, delta } => {
                 self.cluster_state_map.apply_delta(delta);
-                let delta = self.cluster_state_map.compute_delta(&digest, self.max_transmitted_key_values);
+                let delta = self
+                    .cluster_state_map
+                    .compute_delta(&digest, self.max_transmitted_key_values);
                 Some(ScuttleButtMessage::Ack { delta })
             }
             ScuttleButtMessage::Ack { delta } => {
--- a/src/server.rs
+++ b/src/server.rs
@@ -118,9 +118,7 @@ impl UdpServer {
                     }
                     Err(err) => return Err(err.into()),
                 },
-                _ = interval.tick() => {
-                    let _ = self.gossip_multiple(&mut rng);
-                },
+                _ = interval.tick() => self.gossip_multiple(&mut rng).await,
                 message = self.channel.recv() => match message {
                     Some(ChannelMessage::Gossip(addr)) => {
                         let _ = self.gossip(addr).await;
--- a/src/server.rs
+++ b/src/server.rs
@@ -1,8 +1,10 @@
 use std::error::Error;
+use std::sync::Arc;
 
 use tokio::net::UdpSocket;
 use tokio::sync::mpsc::{self, UnboundedReceiver, UnboundedSender};
-use tokio::task::JoinHandle;
+use tokio::sync::Mutex;
+use tokio::task::{JoinError, JoinHandle};
 
 use crate::ScuttleButt;
 
@@ -11,6 +13,8 @@ const BUF_SIZE: usize = 4096;
 
 /// UDP ScuttleButt server.
 pub struct ScuttleServer {
+    pub scuttlebutt: Arc<Mutex<ScuttleButt>>,
+
     channel: UnboundedSender<ChannelMessage>,
     join_handle: JoinHandle<()>,
 }
@@ -19,44 +23,43 @@ impl ScuttleServer {
     /// Launch a new server.
     ///
     /// This will start the ScuttleButt server as a new Tokio background task.
-    pub fn spawn(node_id: impl Into<String>, port: u16) -> Result<Self, Box<dyn Error>> {
+    pub fn spawn(node_id: impl Into<String>, port: u16) -> Self {
         let (tx, rx) = mpsc::unbounded_channel();
-        let node_id = node_id.into();
+
+        let scuttlebutt = Arc::new(Mutex::new(ScuttleButt::with_node_id(node_id.into())));
+        let scuttlebutt_server = scuttlebutt.clone();
 
         let join_handle = tokio::spawn(async move {
-            let _ = listen(rx, node_id, port).await;
+            let _ = listen(rx, scuttlebutt_server, port).await;
         });
 
-        Ok(Self {
+        Self {
             channel: tx,
+            scuttlebutt,
             join_handle,
-        })
+        }
     }
 
     /// Perform a ScuttleButt "handshake" with another UDP server.
-    pub fn gossip(&self, addr: impl Into<String>) -> Result<(), Box<dyn Error>> {
-        self.channel.send(ChannelMessage::Gossip(addr.into()))?;
-        Ok(())
+    pub fn gossip(&self, addr: impl Into<String>) {
+        let _ = self.channel.send(ChannelMessage::Gossip(addr.into()));
     }
 
     /// Shut the server down.
-    pub async fn shutdown(self) -> Result<(), Box<dyn Error>> {
-        self.channel.send(ChannelMessage::Shutdown)?;
-        Ok(self.join_handle.await?)
+    pub async fn shutdown(self) -> Result<(), JoinError> {
+        let _ = self.channel.send(ChannelMessage::Shutdown);
+        self.join_handle.await
     }
 }
 
 /// Listen on the specified UDP port for new ScuttleButt messages.
 async fn listen(
     mut channel: UnboundedReceiver<ChannelMessage>,
-    node_id: String,
+    scuttlebutt: Arc<Mutex<ScuttleButt>>,
     port: u16,
 ) -> Result<(), Box<dyn Error>> {
     let addr = format!("0.0.0.0:{}", port);
     let socket = UdpSocket::bind(addr).await?;
-
-    // TODO: Can't mutate node state after creation.
-    let mut scuttlebutt = ScuttleButt::with_node_id(node_id);
 
     let mut buf = [0; BUF_SIZE];
     loop {
@@ -64,7 +67,7 @@ async fn listen(
             Ok((len, addr)) = socket.recv_from(&mut buf) => {
                 // Handle gossip from other servers.
                 let message = bincode::deserialize(&buf[..len])?;
-                let response = scuttlebutt.process_message(message);
+                let response = scuttlebutt.lock().await.process_message(message);
 
                 // Send reply if necessary.
                 if let Some(message) = response {
@@ -74,7 +77,7 @@ async fn listen(
             },
             Some(message) = channel.recv() => match message {
                 ChannelMessage::Gossip(addr) => {
-                    let syn = scuttlebutt.create_syn_message();
+                    let syn = scuttlebutt.lock().await.create_syn_message();
                     let message = bincode::serialize(&syn)?;
                     let _ = socket.send_to(&message, addr).await?;
                 },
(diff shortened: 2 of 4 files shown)


src/lib.rs
```diff
@@ -11,34 +13,37 @@ pub struct ScuttleButt {
...
+            cluster_state_map: ClusterState::default(),
         }
     }
 
     pub fn set_max_num_key_values(&mut self, max_transmitted_key_values: usize) {
         self.max_transmitted_key_values = max_transmitted_key_values;
     }
 
-    pub fn create_syn_message(&mut self) -> ScuttleButtMessage {
+    pub fn create_syn_message(&self) -> ScuttleButtMessage {
```
chrisduerr, line 28: I've made some minor unrelated changes in this file simply to satisfy clippy/rustfmt. Could revert them if desired but I think improving code quality is always nice.

src/server.rs
```diff
@@ -0,0 +1,145 @@
...
+    /// This will start the ScuttleButt server as a new Tokio background task.
+    pub fn spawn(node_id: impl Into<String>, port: u16) -> Self {
+        let (tx, rx) = mpsc::unbounded_channel();
+
+        let scuttlebutt = Arc::new(Mutex::new(ScuttleButt::with_node_id(node_id.into())));
+        let scuttlebutt_server = scuttlebutt.clone();
+
+        let join_handle = tokio::spawn(async move {
+            let _ = listen(rx, scuttlebutt_server, port).await;
+        });
```
chrisduerr, line 34: One of the main issues I see with just running some background tokio server as discussed in our meeting, is that this is lacking transparency to the consumer. If no tokio runtime is present, the function will fail at runtime.

It would of course be possible to leave it up to the consumer to call the server loop's body in a loop, but that then negatively affects the API surface even though people still need to use a runtime anyway. So I've decided to leave it as-is for now.

src/server.rs
```diff
@@ -0,0 +1,145 @@
...
+
+use tokio::net::UdpSocket;
+use tokio::sync::mpsc::{self, UnboundedReceiver, UnboundedSender};
+use tokio::sync::Mutex;
+use tokio::task::{JoinError, JoinHandle};
+
+use crate::ScuttleButt;
+
+/// Buffer size for UDP messages.
+const BUF_SIZE: usize = 4096;
```
chrisduerr, line 12: This size should be big enough for most messages without having a significant memory impact, but any static buffer size runs risk of getting exceeded since the transmitted data depends on `max_transmitted_key_values` and the size of the k/v sent through the UDP server.

So it might be worth looking into a more dynamically sized buffer that is capable of spanning multiple `recv_from` calls.

chrisduerr (commented):
This implementation probably isn't perfect yet, but due to the minimal issue description I thought it would be best to get some early feedback. It should all work already though and provide a reasonable API.

Due to the nature of this PR, I've decided not to squash any commits, so I can still get on that once this PR is ready.

src/lib.rs
fulmicoton: absolutely.

src/server.rs
fulmicoton: We can eventually change the max kv logic to express it as a buffer len.

src/server.rs
```diff
@@ -0,0 +1,145 @@
...
+        Self {
+            channel: tx,
+            scuttlebutt,
+            join_handle,
+        }
+    }
+
+    /// Perform a ScuttleButt "handshake" with another UDP server.
+    pub fn gossip(&self, addr: impl Into<String>) {
+        let _ = self.channel.send(ChannelMessage::Gossip(addr.into()));
```
fulmicoton, line 45: Can we afford to ignore errors here?

After one kind of trail of events can this return an error and how do we handle that today?

src/server.rs
```diff
@@ -0,0 +1,145 @@
...
+        let mut buf = [0; BUF_SIZE];
+        let (len, _addr) = socket.recv_from(&mut buf).await.unwrap();
+
+        match bincode::deserialize(&buf[..len]).unwrap() {
+            ScuttleButtMessage::SynAck { .. } => (),
+            message => panic!("unexpected message: {:?}", message),
+        }
+
+        server.shutdown().await.unwrap();
+    }
```
fulmicoton, line 144: please add a unit test with a broken payload.

src/server.rs
```diff
@@ -0,0 +1,145 @@
...
+        self.join_handle.await
+    }
+}
+
+/// Listen on the specified UDP port for new ScuttleButt messages.
+async fn listen(
+    mut channel: UnboundedReceiver<ChannelMessage>,
+    scuttlebutt: Arc<Mutex<ScuttleButt>>,
+    port: u16,
+) -> Result<(), Box<dyn Error>> {
```
fulmicoton, line 60:
```suggestion
) -> anyhow::Result<()> {
```

src/server.rs
```diff
@@ -0,0 +1,145 @@
...
+use tokio::task::{JoinError, JoinHandle};
+
+use crate::ScuttleButt;
+
+/// Buffer size for UDP messages.
+const BUF_SIZE: usize = 4096;
+
+/// UDP ScuttleButt server.
+pub struct ScuttleServer {
+    pub scuttlebutt: Arc<Mutex<ScuttleButt>>,
```
fulmicoton, line 16: How does one access the scuttlebutt object?

fulmicoton, line 16: I am a bit scared by the idea of exposing this `Arc<Mutex>` as public.
On the other hand I am not entirely sure how to fix it (copying accessor, per node lock, concurrent hashmap, ...) without us knowing a little more about its usage.

src/server.rs
fulmicoton: Yes this is ok.

This is a quickwit project crate first, not a general purpose library. (at least for the moment)

fulmicoton (changes requested):
See question inline.

I think there is one big flaw. The code does not seem to handle errors well, nor allows the user to deal with errors.

src/server.rs
```diff
@@ -0,0 +1,145 @@
...
+    }
+}
+
+/// Listen on the specified UDP port for new ScuttleButt messages.
+async fn listen(
+    mut channel: UnboundedReceiver<ChannelMessage>,
+    scuttlebutt: Arc<Mutex<ScuttleButt>>,
+    port: u16,
+) -> Result<(), Box<dyn Error>> {
+    let addr = format!("0.0.0.0:{}", port);
```
fulmicoton, line 61: It is difficult to explain that at this point, but I think we actually do not want to listen at all network interfaces.

src/server.rs
fulmicoton:
```suggestion
/// Buffer size for UDP messages.
/// This value is set to avoid UDP fragmentation.
/// https://jvns.ca/blog/2017/02/07/mtu/
const BUF_SIZE: usize = 1_472;
```

fulmicoton: I'll try to scuttlebutt model part to include serialization and express mtu in bytes.

fulmicoton (changes requested):
See comment inline

src/server.rs
fulmicoton: Actually can we remove the fixed-size stack buffer the moment and simply use a `Vec` at least for the moment?
The max number of tuple logic should be enough most of the time, and fragment UDP packets are supposed to deal with the rest.

src/server.rs
chrisduerr: >  After one kind of trail of events can this return an error and how do we handle that today?

I'm not sure what you mean by "trail of events" here. This should only return an error when the server has shut down and the receiver was dropped. It's probably useful to return a Result from it just in case.

src/server.rs
```diff
@@ -0,0 +1,294 @@
...
+    /// Listen for new ScuttleButt messages.
+    async fn run(&mut self) -> anyhow::Result<()> {
+        let mut buf = [0; BUF_SIZE];
+        loop {
+            tokio::select! {
+                result = self.socket.recv_from(&mut buf) => match result {
+                    Ok((len, addr)) => {
+                        let _ = self.process_package(addr, &buf[..len]).await;
+                    }
+                    Err(err) => return Err(err.into()),
```
chrisduerr, line 124: A payload of excessive size is ignored, I've chosen to explicitly add a test for that. But this is still necessary in case something went wrong with the socket. I don't think there's much that can be done as a recovery from this specific error.

src/server.rs
```diff
@@ -0,0 +1,294 @@
...
+                },
+                message = self.channel.recv() => match message {
+                    Some(ChannelMessage::Gossip(addr)) => {
+                        let _ = self.gossip(addr).await;
+                    },
+                    Some(ChannelMessage::Shutdown) | None => break,
```
chrisduerr, line 130: The client channel dropping is now handled the same way as an explicit shutdown. That way the server doesn't stick around when nobody needs it anymore.

src/server.rs
```diff
@@ -0,0 +1,294 @@
...
+    use std::future::Future;
+    use std::time::Duration;
+
+    use crate::ScuttleButtMessage;
+
+    async fn timeout<O>(future: impl Future<Output = O>) -> O {
+        tokio::time::timeout(Duration::from_millis(100), future)
+            .await
+            .unwrap()
+    }
```
chrisduerr, line 181: The timeout is never needed in cases where the test passes, but I think it's a good idea to make failure quick for development.

chrisduerr:
Unfortunately I was sick over the weekend/last week, but I've found some time on Monday to address the review feedback. I think I should have gotten all the big things out of the way.

src/server.rs
```diff
@@ -0,0 +1,268 @@
...
+use tokio::sync::mpsc::{self, UnboundedReceiver, UnboundedSender};
+use tokio::sync::Mutex;
+use tokio::task::JoinHandle;
+
+use crate::ScuttleButt;
+
+/// Buffer size for UDP messages.
+/// This value is set to avoid UDP fragmentation.
+/// https://jvns.ca/blog/2017/02/07/mtu/
+const BUF_SIZE: usize = 1_472;
```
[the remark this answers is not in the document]
fulmicoton: Ethernet mtu is usually 1500, which leads to a payload of 1472 for UDP.

I think you are right. The story on AWS is quite more complicated.
Let's go for 1024.
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/network_mtu.html

src/server.rs
```diff
@@ -0,0 +1,294 @@
...
+    /// This will automatically gossip about the changes to all known nodes.
+    pub async fn with_scuttlebutt_mut<F, T>(&self, mut fun: F) -> T
+    where
+        F: FnMut(&mut ScuttleButt) -> T,
+    {
+        let mut scuttlebutt = self.scuttlebutt.lock().await;
+        let retval = fun(&mut scuttlebutt);
+
+        // Broadcast changes to all nodes after mutation.
+        for node in scuttlebutt.cluster_state_map.nodes() {
```
fulmicoton, line 72: Gossiping to all nodes would defy the purpose.

The point of that kind of gossip algorithm is to keep a bounded budget of network use.
We want a task that triggers a gossip round periodically (every second or so). 

Gossipping here means:
- picking 3 nodes randomly
- gossipping to them

You can check the cassandra links to see their refinements.

a) They pick from the alive node pool first.
b) They pick from a special seed + alive pool.
c) Sometimes they try to contact a dead node

We can pick nodes at random for the moment as we do not have a notion of liveness yet.
The algorithm will still be correct.

src/server.rs
```diff
@@ -0,0 +1,268 @@
...
+    channel: UnboundedSender<ChannelMessage>,
+    scuttlebutt: Arc<Mutex<ScuttleButt>>,
+    join_handle: JoinHandle<Result<(), anyhow::Error>>,
+}
+
+impl ScuttleServer {
+    /// Launch a new server.
+    ///
+    /// This will start the ScuttleButt server as a new Tokio background task.
+    pub fn spawn(address: impl Into<String>, seed_nodes: Vec<String>) -> Self {
```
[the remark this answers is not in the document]
fulmicoton: sounds good to me.

src/server.rs
```diff
@@ -0,0 +1,294 @@
...
+
+    /// Shut the server down.
+    pub async fn shutdown(self) -> Result<(), anyhow::Error> {
+        let _ = self.channel.send(ChannelMessage::Shutdown);
+        self.join_handle.await?
+    }
+
+    /// Perform a ScuttleButt "handshake" with another UDP server.
+    pub fn gossip(&self, addr: impl Into<String>) -> Result<(), anyhow::Error> {
+        Ok(self.channel.send(ChannelMessage::Gossip(addr.into()))?)
```
fulmicoton, line 87:
```suggestion
    pub fn gossip(&self, addr: impl Into<String>) -> anyhow::Result<()> {
        self.channel.send(ChannelMessage::Gossip(addr.into()))?;
        Ok(())
```

`.context()` could be nice too.

src/server.rs
```diff
@@ -0,0 +1,268 @@
...
+    }
```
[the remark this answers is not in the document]
fulmicoton: See above. The with_scuttlebutt_mut does not need to perform gossip.
Gossipping is a thing that happens always, as a background task.

src/server.rs
```diff
@@ -0,0 +1,268 @@
...
+            socket,
+        })
+    }
+
+    /// Listen for new ScuttleButt messages.
+    async fn run(&mut self) -> anyhow::Result<()> {
+        let mut buf = [0; BUF_SIZE];
+        loop {
+            tokio::select! {
+                result = self.socket.recv_from(&mut buf) => match result {
```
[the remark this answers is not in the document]
fulmicoton: I think ignoring them wrong message might be fine. logging a warning of something might be a good idea though.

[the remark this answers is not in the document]
fulmicoton: Yeah the program itself will try to never exceed `BUF_SIZE`. I am more scared on anythign that could send random stuff on the security port. A security audit, etc.

src/server.rs
fulmicoton: I think this is good.

src/server.rs
fulmicoton: That makes sense.

src/server.rs
```diff
@@ -0,0 +1,268 @@
...
+        F: FnMut(&ScuttleButt) -> T,
+    {
+        let scuttlebutt = self.scuttlebutt.lock().await;
+        fun(&scuttlebutt)
+    }
+
+    /// Call a function with mutable access to the [`ScuttleButt`].
+    ///
+    /// This will automatically gossip about the changes to all known nodes.
+    pub async fn with_scuttlebutt_mut<F, T>(&self, mut fun: F) -> T
```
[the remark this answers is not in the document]
fulmicoton: I used the same pattern very recently. 
I like that it prevents exposing the locking.

fulmicoton:
@chrisduerr  I think we can merge it and maybe work on another PR.
Look at the comment though.

I think next step is to spawn a tasks that gossips periodically to a subset of random nodes.

src/server.rs
fulmicoton: we are within that map today. It might not be a good idea to send a gossip message to ourselves.

fulmicoton:
@chrisduerr please have a look at the discussion and do any change you might want to do. When you think we are ready to merge we can merge and continue in another PR.

src/server.rs
chrisduerr: > Gossiping to all nodes would defy the purpose.

That actually makes a lot of sense, yeah. I wasn't aware of the exact behavior but I think my latest commit should work much better.

In that case the whole `with_` stuff wouldn't really be necessary anymore. However this still "hides" the mutex lock. Do you prefer a `with_` method over exposing the `Arc<Mutex<_>>` publicly (or a `MutexGuard<_>` I suppose)?

I don't think I do, but you made it sound like you're quite opposed to having the lock interface public.

src/server.rs
chrisduerr: I'm not familiar with `anyhow` or its `context` method, but I feel like any handling of this error should make the source quite obvious? If you have specific information you'd like to attach in the context I'm open to adding it though.

fulmicoton: ok

src/server.rs
fulmicoton: I like the idea of passing a function.

PR #5, opened 2021-12-21
Add a heartbeat to the UDP server

### LINKED ISSUE #3: Introduce a concept of heartbeat.

chrisduerr:
Fixes #3.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -66,6 +66,10 @@ impl ScuttleButt {
 mod tests {
     use super::*;
 
+    use std::collections::HashMap;
+    use std::hash::Hash;
+    use std::fmt::Debug;
+
     fn run_scuttlebutt_handshake(initiating_node: &mut ScuttleButt, peer_node: &mut ScuttleButt) {
         let syn_message = initiating_node.create_syn_message();
         let syn_ack_message = peer_node.process_message(syn_message).unwrap();
@@ -73,16 +77,27 @@ mod tests {
         assert!(peer_node.process_message(ack_message).is_none());
     }
 
-    fn assert_nodes_sync(nodes: &[&ScuttleButt]) -> bool {
-        if nodes.len() <= 1 {
-            return true;
+    fn assert_map_eq<K, V>(lhs: &HashMap<K, V>, rhs: &HashMap<K, V>)
+    where
+        K: Eq + Hash,
+        V: PartialEq + Debug,
+    {
+        assert_eq!(lhs.len(), rhs.len());
+        for (key, value) in lhs {
+            assert_eq!(rhs.get(key), Some(value));
         }
-        let first_node_state = serde_json::to_value(&nodes[0].cluster_state_map).unwrap();
+    }
+
+    fn assert_nodes_sync(nodes: &[&ScuttleButt]) {
+        let first_node_states = &nodes[0].cluster_state_map.node_states;
         for other_node in nodes.iter().skip(1) {
-            let node_state = serde_json::to_value(&other_node.cluster_state_map).unwrap();
-            assert_json_diff::assert_json_eq!(&first_node_state, &node_state);
+            let node_states = &other_node.cluster_state_map.node_states;
+
+            assert_eq!(first_node_states.len(), node_states.len());
+            for (key, value) in first_node_states {
+                assert_map_eq(&value.key_values, &node_states.get(key).unwrap().key_values);
+            }
         }
-        false
     }
 
     #[test]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2,7 +2,7 @@ pub mod server;
 
 mod model;
 
-use crate::model::{ClusterState, NodeState, ScuttleButtMessage};
+use crate::model::{ClusterState, NodeState, ScuttleButtMessage, Digest};
 
 // https://www.cs.cornell.edu/home/rvr/papers/flowgossip.pdf
 
@@ -36,11 +36,7 @@ impl ScuttleButt {
     }
 
     pub fn create_syn_message(&mut self) -> ScuttleButtMessage {
-        self.heartbeat += 1;
-        let heartbeat = self.heartbeat;
-        self.self_node_state().set(HEARTBEAT_KEY, heartbeat);
-
-        let digest = self.cluster_state_map.compute_digest();
+        let digest = self.compute_digest();
         ScuttleButtMessage::Syn { digest }
     }
 
@@ -50,7 +46,7 @@ impl ScuttleButt {
                 let delta = self
                     .cluster_state_map
                     .compute_delta(&digest, self.max_transmitted_key_values);
-                let digest = self.cluster_state_map.compute_digest();
+                let digest = self.compute_digest();
                 Some(ScuttleButtMessage::SynAck { delta, digest })
             }
             ScuttleButtMessage::SynAck { digest, delta } => {
@@ -78,6 +74,16 @@ impl ScuttleButt {
     /// Retrieve a list of all living nodes.
     pub fn living_nodes(&self) -> impl Iterator<Item = &str> {
         self.cluster_state_map.living_nodes()
+    }
+
+    /// Compute digest, always at least adding a heartbeat as update.
+    fn compute_digest(&self) -> Digest {
+        // Ensure for every reply from this node, at least the heartbeat is changed.
+        self.heartbeat += 1;
+        let heartbeat = self.heartbeat;
+        self.self_node_state().set(HEARTBEAT_KEY, heartbeat);
+
+        self.cluster_state_map.compute_digest()
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6,26 +6,40 @@ use crate::model::{ClusterState, NodeState, ScuttleButtMessage};
 
 // https://www.cs.cornell.edu/home/rvr/papers/flowgossip.pdf
 
+/// HashMap key for the heartbeat node value.
+pub(crate) const HEARTBEAT_KEY: &str = "heartbeat";
+
 pub struct ScuttleButt {
     max_transmitted_key_values: usize, // mtu in the paper
     self_node_id: String,
     cluster_state_map: ClusterState,
+    heartbeat: u64,
 }
 
 impl ScuttleButt {
     pub fn with_node_id(self_node_id: String) -> Self {
-        ScuttleButt {
+        let mut scuttlebutt = ScuttleButt {
             max_transmitted_key_values: 10,
+            heartbeat: 0,
             self_node_id,
             cluster_state_map: ClusterState::default(),
-        }
+        };
+
+        // Immediately mark node as alive to ensure it responds to SYNs.
+        scuttlebutt.self_node_state().set(HEARTBEAT_KEY, 0);
+
+        scuttlebutt
     }
 
     pub fn set_max_num_key_values(&mut self, max_transmitted_key_values: usize) {
         self.max_transmitted_key_values = max_transmitted_key_values;
     }
 
-    pub fn create_syn_message(&self) -> ScuttleButtMessage {
+    pub fn create_syn_message(&mut self) -> ScuttleButtMessage {
+        self.heartbeat += 1;
+        let heartbeat = self.heartbeat;
+        self.self_node_state().set(HEARTBEAT_KEY, heartbeat);
+
         let digest = self.cluster_state_map.compute_digest();
         ScuttleButtMessage::Syn { digest }
     }
@@ -60,6 +74,11 @@ impl ScuttleButt {
     pub fn self_node_state(&mut self) -> &mut NodeState {
         self.cluster_state_map.node_state_mut(&self.self_node_id)
     }
+
+    /// Retrieve a list of all living nodes.
+    pub fn living_nodes(&self) -> impl Iterator<Item = &str> {
+        self.cluster_state_map.living_nodes()
+    }
 }
 
 #[cfg(test)]
@@ -67,8 +86,8 @@ mod tests {
     use super::*;
 
     use std::collections::HashMap;
-    use std::hash::Hash;
     use std::fmt::Debug;
+    use std::hash::Hash;
 
     fn run_scuttlebutt_handshake(initiating_node: &mut ScuttleButt, peer_node: &mut ScuttleButt) {
         let syn_message = initiating_node.create_syn_message();
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -76,7 +76,11 @@ impl ScuttleButt {
         self.cluster_state_map.living_nodes()
     }
 
-    /// Compute digest, always at least adding a heartbeat as update.
+    /// Compute digest.
+    ///
+    /// This method also increments the heartbeat, to force the presence
+    /// of at least one update, and have the node liveliness propagated
+    /// through the cluster.
     fn compute_digest(&self) -> Digest {
         // Ensure for every reply from this node, at least the heartbeat is changed.
         self.heartbeat += 1;
--- a/src/model.rs
+++ b/src/model.rs
@@ -159,10 +159,10 @@ impl ClusterState {
         self.node_states.get(node_id)
     }
 
-    // TODO: Just a demo.
-    pub fn living_nodes(&self) -> impl Iterator<Item = &String> {
+    /// Retrieve a list of all living nodes.
+    pub fn living_nodes(&self) -> impl Iterator<Item = &str> {
         self.node_states.iter().filter_map(|(node_id, node_state)| {
-            (node_state.last_heartbeat.elapsed() <= MAX_HEARTBEAT_DELTA).then(|| node_id)
+            (node_state.last_heartbeat.elapsed() <= MAX_HEARTBEAT_DELTA).then(|| node_id.as_str())
         })
     }
 
--- a/src/model.rs
+++ b/src/model.rs
@@ -149,6 +149,10 @@ impl ClusterState {
         self.node_states.keys()
     }
 
+    pub fn node_states(&mut self) -> &mut HashMap<String, NodeState> {
+        &mut self.node_states
+    }
+
     pub fn apply_delta(&mut self, delta: Delta) {
         for (node_id, node_delta) in delta.node_deltas {
             let mut node_state_map = self
--- a/src/model.rs
+++ b/src/model.rs
@@ -89,9 +89,9 @@ pub enum ScuttleButtMessage {
 }
 
 pub struct NodeState {
-    max_version: u64,
-    key_values: HashMap<String, VersionedValue>,
+    pub(crate) key_values: HashMap<String, VersionedValue>,
     last_heartbeat: Instant,
+    max_version: u64,
 }
 
 impl Default for NodeState {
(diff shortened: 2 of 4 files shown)


src/server.rs
```diff
@@ -0,0 +1,346 @@
...
+        let mut scuttlebutt = self.scuttlebutt.lock().await;
+        let nodes = scuttlebutt.cluster_state_map.node_states();
+        let now = Self::secs_since_epoch();
+
+        nodes.retain(|_node_id, node_state| {
+            let heartbeat = node_state
+                .get(HEARTBEAT_KEY)
+                .and_then(|value| value.parse().ok());
+            now - heartbeat.unwrap_or(0) <= MAX_HEARTBEAT_DELTA.as_secs()
+        });
```
chrisduerr, line 167: This is what I'd consider a "naive" approach to heartbeats: Remove all nodes that exceed a specific duration.

However, looking at the `compute_delta` method it seems like this could behave rather strangely with gossiping, since nodes aren't removed on gossip. So every node would have to remove it themselves and it would be possible that one node removes the node, another gossips about it because its timer runs slightly differently and then the original node suddenly has the dead node again.

Sooner or later all nodes will remove this themselves, but I'm not sure if a "proper" solution wouldn't filter dead nodes *before* gossiping in addition to periodically? Or just before gossiping in general since that happens periodically? I haven't read through the Cassandra source code, but I'd be surprised if they just roll with this issue.

chrisduerr (commented):
I had a second of time so I decided to check how a heartbeat might be implemented. But this is mostly a patch to demonstrate my question about heartbeats rather than a proper implementation.

src/server.rs
```diff
@@ -0,0 +1,346 @@
...
+    async fn reap_zombies(&self) {
+        let mut scuttlebutt = self.scuttlebutt.lock().await;
+        let nodes = scuttlebutt.cluster_state_map.node_states();
+        let now = Self::secs_since_epoch();
+
+        nodes.retain(|_node_id, node_state| {
+            let heartbeat = node_state
+                .get(HEARTBEAT_KEY)
+                .and_then(|value| value.parse().ok());
+            now - heartbeat.unwrap_or(0) <= MAX_HEARTBEAT_DELTA.as_secs()
```
fulmicoton, line 166: No. You want to use a monotonous clock to compute timeout.

src/server.rs
```diff
@@ -0,0 +1,346 @@
...
+            let _ = self.gossip(*node).await;
+        }
+    }
+
+    /// Gossip to one other UDP server.
+    async fn gossip(&self, addr: impl Into<String>) -> anyhow::Result<()> {
+        let mut scuttlebutt = self.scuttlebutt.lock().await;
+        scuttlebutt
+            .self_node_state()
+            .set(HEARTBEAT_KEY, Self::secs_since_epoch());
```
fulmicoton, line 193: No. You are assuming nodes have a clock that in sync here.

src/server.rs
fulmicoton: We never remove nodes.

Each just separately assesses whether they think a node is dead or not. 
This information is not shared with other nodes, but an internal client in the node can get the list of live nodes.

src/server.rs
fulmicoton: The solution for b) is tricky of course.

The idea is to just use a autoincrement counter as the heartbeat.

The counter in itself is not used.

For a node N to track the liveliness of a given node N' we do need some kind of time information though. 
We get it by calling `Instant::now()` and associating the value to N' whenever we get a sign of liveliness from it:
e.g.:
a) we (the N node) received a message from the N' node
b) we (the N node) received an update about N' from another node N''

Note that in the case of b) the update information could be quite lagging. This is not a problem that hurts the validity of the algorithm. A dead will still be eventually detected as dead.

fulmicoton (changes requested):
See comments inline.

The most important part  is that wall time is useless here, and time in general cannot be shared between nodes.

Working with a threshold is ok for the moment. Phi accrual detection is all about making this threshold dynamic.

src/server.rs
chrisduerr: Ah okay, this probably answers a question I forgot to pose: It's not necessary to have any push-based API informing consumers about node deaths, right? So the consumers of the API just query for which nodes are alive whenever they need that information.

src/server.rs
chrisduerr: So I'd assume the heartbeat's sole purpose then is to mark a node as "dirty" so nodes will receive an update and update their liveliness status of that node? It's never actually consumed?

And the liveliness tracked in N for N' isn't stored on the N'/Ns k/v storage, but externally, so we don't have that information constantly creating more gossip and create potential feedback-loops?

fulmicoton: > And the liveliness tracked in N for N' isn't stored on the N'/Ns k/v storage

Yes we don't store that in the gossiped storage.

> It's never actually consumed?

Within the same process, we need a method to list the "live" nodes.

src/server.rs
fulmicoton: That's a good question.
We can do a pull API only for the moment I think.

src/server.rs
chrisduerr: > Within the same process, we need a method to list the "live" nodes.

I was talking about the heartbeat stored in the k/v storage in this scenario. That's never "consumed" by any other nodes as far as I understood it.

src/model.rs
```diff
@@ -130,7 +144,7 @@ impl NodeState {
     }
 }
 
-#[derive(Default, Serialize)]
+#[derive(Default)]
```
chrisduerr, line 147: I've removed `Serialize` here because `Instant` doesn't trivially implement it. It wasn't used anywhere so I'd be curious if this is necessary for consumption of the API?

Since it's not trivial to just convert an Instant to seconds or something like that (at least with STD) that would likely make things at least a tiny bit more complicated. Worst-case scenario one might have to pass in something like a startup time stored on the `ScuttleButt`, which doesn't need Serialize, that would definitely work. But maybe not necessary?

src/model.rs
```diff
@@ -84,10 +88,20 @@ pub enum ScuttleButtMessage {
     Ack { delta: Delta },
 }
 
-#[derive(Default, Serialize)]
 pub struct NodeState {
     max_version: u64,
     key_values: HashMap<String, VersionedValue>,
+    last_heartbeat: Instant,
```
chrisduerr, line 94: If I understand phi accrual failure detection correctly, all one would have to do to implement it with my current approach is store an array of Instants here. Then it should be trivial to compute expected heartbeat arrival time in the `living_nodes` function.

src/model.rs
```diff
@@ -145,6 +159,13 @@ impl ClusterState {
         self.node_states.get(node_id)
     }
 
+    // TODO: Just a demo.
```
chrisduerr, line 162: Currently there is no way to consume this API. There is no way to access all nodes from the outside at all (just a specific node, but how would a consumer know about it to begin with?). I'd assume the information one would care about externally is the `node_id` and k/v of all nodes (so basically everything). In that case one could probably just expose this method publicly through `ScuttleButt`? Using the `node_id`s then the consumer can access the k/v of all nodes, alternatively this method could expose k/v and node_id at the same time, since it wouldn't really be any additional effort anyway.

src/server.rs
```diff
@@ -165,7 +168,17 @@ impl UdpServer {
...
-        let syn = self.scuttlebutt.lock().await.create_syn_message();
+        let mut scuttlebutt = self.scuttlebutt.lock().await;
+        let heartbeat = scuttlebutt
+            .self_node_state()
+            .get(HEARTBEAT_KEY)
+            .and_then(|heartbeat| str::parse(heartbeat).ok())
+            .unwrap_or(0);
+        scuttlebutt
+            .self_node_state()
+            .set(HEARTBEAT_KEY, heartbeat + 1);
```
chrisduerr, line 179: I'd say parsing a string here is pretty ugly, but I'd assume the functionality is mostly what you were thinking of?

src/server.rs
fulmicoton: Ah yes. The value is not consumed. The only interesting signal is that it changed.

src/model.rs
fulmicoton: This last_heartbeat thing is not gossiped and lives outside of the Scuttlebutt state. It does not need to be serialized.

src/model.rs
fulmicoton: Yes I think it is a good idea to have this method.

I recommend &str instead of &String however.
(Granted, it does not matter as much as a return value)

src/server.rs
fulmicoton: Yeah let's not parse it.
What about having a local heartbeat in the object and auto increment it?
We can make it an atomic if it helps.

src/server.rs
```diff
@@ -165,7 +168,17 @@ impl UdpServer {
...
     async fn gossip(&self, addr: impl Into<String>) -> anyhow::Result<()> {
-        let syn = self.scuttlebutt.lock().await.create_syn_message();
+        let mut scuttlebutt = self.scuttlebutt.lock().await;
+        let heartbeat = scuttlebutt
+            .self_node_state()
+            .get(HEARTBEAT_KEY)
+            .and_then(|heartbeat| str::parse(heartbeat).ok())
+            .unwrap_or(0);
+        scuttlebutt
+            .self_node_state()
```
fulmicoton, line 178: You called self_node_state twice.
That thing takes a lock. That's a race condition here.
If you call gossip concurrently twice, the value for the heartbeat is not well-defined.

src/model.rs
fulmicoton: I believe this is correct.

src/server.rs
chrisduerr: I'm probably missing something, but the lock is held across both get and set of `self_node_state`. I don't see any locks in the model/lib so I'd assume you're talking about the `scuttlebutt` lock here?

In that scenario this shouldn't be any different than `let hb = scuttlebutt.hb + 1; scuttlebutt.hb = hb;` unless there's some chance for async Rust to put tasks into the background without any `.await` points in between (which I don't think it is? Admittedly not particularly experienced with async Rust).

fulmicoton: Ah yes my bad! I thought self node state was taking a lock and returning a mutexguard. Sorry I confused myself.

src/lib.rs
```diff
@@ -6,26 +6,40 @@ use crate::model::{ClusterState, NodeState, ScuttleButtMessage};
...
+        let mut scuttlebutt = ScuttleButt {
             max_transmitted_key_values: 10,
+            heartbeat: 0,
             self_node_id,
             cluster_state_map: ClusterState::default(),
-        }
+        };
+
+        // Immediately mark node as alive to ensure it responds to SYNs.
+        scuttlebutt.self_node_state().set(HEARTBEAT_KEY, 0);
```
chrisduerr, line 29: Had to add this here because when writing my test I found out that if the node is completely empty, it will not be added to the cluster state and thus not be listed as a living node. I'd assume we want nodes to be alive from the get-go if they receive a syn before sending one on their own.

That said, as long as there are seed nodes usually a `ScuttleButt` will never be empty anyway.

src/lib.rs
```diff
@@ -6,26 +6,40 @@ use crate::model::{ClusterState, NodeState, ScuttleButtMessage};
...
 
     pub fn set_max_num_key_values(&mut self, max_transmitted_key_values: usize) {
         self.max_transmitted_key_values = max_transmitted_key_values;
     }
 
-    pub fn create_syn_message(&self) -> ScuttleButtMessage {
+    pub fn create_syn_message(&mut self) -> ScuttleButtMessage {
+        self.heartbeat += 1;
+        let heartbeat = self.heartbeat;
+        self.self_node_state().set(HEARTBEAT_KEY, heartbeat);
```
chrisduerr, line 41: This is where the heartbeat lives now. I think having it on the `create_syn` makes sense to automagically do it for the consumer regardless of using a server or not.

Thought about adding it to the syn_ack/ack replies too but I feel like a heartbeat shouldn't be incremented when the node isn't capable of creating syns itself anymore.

src/lib.rs
```diff
@@ -60,29 +74,49 @@ impl ScuttleButt {
     pub fn self_node_state(&mut self) -> &mut NodeState {
         self.cluster_state_map.node_state_mut(&self.self_node_id)
     }
+
+    /// Retrieve a list of all living nodes.
+    pub fn living_nodes(&self) -> impl Iterator<Item = &str> {
+        self.cluster_state_map.living_nodes()
+    }
```
chrisduerr, line 81: This function is now also part of the public API.

src/lib.rs
```diff
@@ -60,29 +74,49 @@ impl ScuttleButt {
...
+        let first_node_states = &nodes[0].cluster_state_map.node_states;
         for other_node in nodes.iter().skip(1) {
-            let node_state = serde_json::to_value(&other_node.cluster_state_map).unwrap();
-            assert_json_diff::assert_json_eq!(&first_node_state, &node_state);
+            let node_states = &other_node.cluster_state_map.node_states;
+
+            assert_eq!(first_node_states.len(), node_states.len());
+            for (key, value) in first_node_states {
+                assert_map_eq(&value.key_values, &node_states.get(key).unwrap().key_values);
+            }
```
chrisduerr, line 118: This should be a pretty simple replacement for the json conversion without requiring serde. Only tests the actual K/V pairs instead of the entire node state, but I think that's the most important part here and the verbosity of testing every single field isn't beneficial.

src/server.rs
```diff
@@ -149,23 +149,28 @@ impl UdpServer {
...
         // Select up to [`GOSSIP_COUNT`] node IDs at random.
         let nodes = scuttlebutt.cluster_state_map.nodes();
         let count = nodes
             .filter(|node_id| node_id != &self_node_id)
-            .map(|node_id| node_id.as_str())
+            .cloned()
             .choose_multiple_fill(rng, &mut rand_nodes);
 
+        // Drop lock to prevent deadlock in [`UdpSocket::gossip`].
+        drop(scuttlebutt);
```
chrisduerr, line 163: Now this is kinda interesting. After you mentioned potential locking issues I immediately added a comment to ensure that the lock created in `gossip_multiple` doesn't create issues with the lock in `gossip`.

Reading the tokio documentation it sounded like holding a lock across await boundaries wasn't that big of a deal, since that seems to be its main benefit over STD's lock. Unfortunately the behavior isn't really documented precisely, but after writing the `heartbeat` test I found out that this **will** cause a deadlock.

So unfortunately this allocates strings now, but at least it doesn't deadlock which is *probably* more important.

src/server.rs
```diff
@@ -308,4 +313,49 @@ mod tests {
 
         server.shutdown().await.unwrap();
     }
+
+    #[tokio::test]
+    async fn heartbeat() {
```
chrisduerr, line 318: This is a pretty big test, but it basically tests an entire lifecycle of the scuttlebutt server. So that should give at least some confidence things actually work properly now.

chrisduerr (commented):
Feel much better about the patch now, especially with the deadlock gone and a test making sure things actually work. Let me know if you have any more ideas for improvement.

src/lib.rs
fulmicoton: No we want the heartbeat to change on replies too.

(random thought: we could make it into a Lamport clock actually?)

src/lib.rs
fulmicoton: yes

src/server.rs
fulmicoton: yes. I am not against that kind of tests.

fulmicoton (changes requested):
Can we update the nodes when sending a message?

src/lib.rs
chrisduerr: > We want the heartbeat to change on replies too.

Sure, that's an easy change.

> (random thought: we could make it into a Lamport clock actually?)

I'm curious what you'd use a Lamport clock for here. IIRC the K/V storage is already versioned and the heartbeat is only ever produced by one node and never consumed by another. So the thing that matters for a heartbeat is the duration since the last time a node was confirmed to be alive, while the actual order of heartbeats doesn't *really* matter.

The only thing I could see is that you'd be able to discard heartbeats that for some reason got severely delayed by the gossiping, which have been produced *after* heartbeats you've already received? But since dead nodes not produce new events, I doubt that would cause them to stick around for **much** longer and they'll eventually be detected as dead anyway.

src/lib.rs
```diff
@@ -60,29 +70,59 @@ impl ScuttleButt {
...
+    }
+
+    /// Compute digest, always at least adding a heartbeat as update.
+    fn compute_digest(&self) -> Digest {
+        // Ensure for every reply from this node, at least the heartbeat is changed.
+        self.heartbeat += 1;
+        let heartbeat = self.heartbeat;
+        self.self_node_state().set(HEARTBEAT_KEY, heartbeat);
+
+        self.cluster_state_map.compute_digest()
```
chrisduerr, line 86: Added this `compute_digest` method here, which can be used to increment the heartbeat and compute the digest of the rest of the cluster.

That way whenever we compute the digest to check if there's anything that changed, *at the very least* the heartbeat will be a delta of the current node. That should hopefully make this pretty automagic even with potential future changes (assuming nobody accidentally calls the `compute_digest` method directly).

I've considered adding it to the `compute_digest` method in the `ClusterState`, but that doesn't have any notion of its own node and I didn't really like the idea of introducing it. This way the cluster state stays as a generic state for the entire cluster, regardless of where we fall within it.

src/lib.rs
fulmicoton: We have no usage for lamport clocks :)

It's just that we need a heartbeat and it can be any value that changes.

- local autoincrement id
- seconds elapsed since startup time
- lamport clock

The lamport clocks might be nice when we debug.
It will make the heartbeat clock comparable, and therefore, make it possible to identify which node is lagging.

On the other hand, seconds elapsed since startup time, is also valuable information for debug.

chrisduerr: Happy to go with whatever you think fits best here, assuming it doesn't involve a mountain of additional code.

Though there is something to be said about just using the simplest solution possible when it comes to debugging. The easiest bug to fix is the one that was prevented before it was made after all :smile:. Not that any of this code itself is particularly likely to cause troubles.

fulmicoton: @chrisduerr You are right. Let's stick to seconds elapsed since startup time.

src/lib.rs
fulmicoton: I think that should do the job for the moment.

src/lib.rs
```diff
@@ -60,29 +70,59 @@ impl ScuttleButt {
     pub fn self_node_state(&mut self) -> &mut NodeState {
         self.cluster_state_map.node_state_mut(&self.self_node_id)
     }
+
+    /// Retrieve a list of all living nodes.
+    pub fn living_nodes(&self) -> impl Iterator<Item = &str> {
+        self.cluster_state_map.living_nodes()
+    }
+
+    /// Compute digest, always at least adding a heartbeat as update.
```
fulmicoton, line 79:
```suggestion
    /// Compute digest.
    ///
    /// This method also increments the heartbeat, to force the presence
    /// of at least one update, and have the node liveliness propagated
    /// through the cluster.
```

fulmicoton (approved):

Issue #11, opened 2022-01-10
Better peer selection and lifecycle


fulmicoton:
Right now we pick peers at random amongst all peers... We can probably do a bit better. 
Also avoid the network partition caused by number of seeds https://issues.apache.org/jira/browse/CASSANDRA-150

PR #15, opened 2022-02-07
Better peer selection

evanxg852000:
- Fixed peer selection logic.
- Guard against network partition

https://github.com/quickwit-oss/scuttlebutt/issues/11

### DIFF
--- a/scuttlebutt-test/Cargo.toml
+++ b/scuttlebutt-test/Cargo.toml
@@ -11,6 +11,7 @@ poem = "1"
 poem-openapi = {version="1.2", features = ["swagger-ui"] }
 structopt = "0.3"
 tokio = { version = "1.14.0", features = ["net", "sync", "rt-multi-thread", "macros", "time"] }
+serde = { version="1", features=["derive"] }
 serde_json = "1"
 anyhow = "1"
 once_cell = "1"
@@ -21,4 +22,4 @@ tracing-subscriber = "0.3"
 assert_cmd = "2"
 predicates = "2"
 reqwest = { version = "0.11", default-features=false, features = ["blocking", "json"] }
-
+serial_test = "0.5.1"
--- a/scuttlebutt-test/Cargo.toml
+++ b/scuttlebutt-test/Cargo.toml
@@ -22,4 +22,3 @@ tracing-subscriber = "0.3"
 assert_cmd = "2"
 predicates = "2"
 reqwest = { version = "0.11", default-features=false, features = ["blocking", "json"] }
-serial_test = "0.5.1"
--- a/scuttlebutt/src/failure_detector.rs
+++ b/scuttlebutt/src/failure_detector.rs
@@ -38,6 +38,8 @@ pub struct FailureDetector {
     config: FailureDetectorConfig,
     /// Denotes live nodes.
     live_nodes: HashSet<NodeId>,
+    /// Denotes dead nodes.
+    dead_nodes: HashSet<NodeId>,
 }
 
 impl FailureDetector {
@@ -46,6 +48,7 @@ impl FailureDetector {
             node_samples: RwLock::new(HashMap::new()),
             config,
             live_nodes: HashSet::new(),
+            dead_nodes: HashSet::new(),
         }
     }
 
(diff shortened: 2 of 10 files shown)


scuttlebutt/src/server.rs
```diff
@@ -172,30 +173,113 @@ impl UdpServer {
...
         const EMPTY_STRING: String = String::new();
         let mut rand_nodes = [EMPTY_STRING; GOSSIP_COUNT];
 
-        // Select up to [`GOSSIP_COUNT`] node IDs at random.
-        let nodes = scuttlebutt_guard.cluster_state.nodes();
-        let count = nodes
-            .filter(|node_id| node_id != self_node_id)
+        // Select `GOSSIP_COUNT` number of live nodes.
+        // On startup, select from cluster nodes since we don't know any live node yet.
+        let count = if live_nodes_count == 0 {
```
guilload, line 185: Would it be simpler to consider the seed nodes live on startup?

scuttlebutt/src/failure_detector.rs
```diff
@@ -69,20 +72,27 @@ impl FailureDetector {
...
             } else {
                 self.live_nodes.insert(node_id.to_string());
+                self.dead_nodes.remove(node_id);
             }
         }
     }
 
-    /// Returns a list of living nodes.
+    /// Returns a list of live nodes.
     pub fn live_nodes(&self) -> impl Iterator<Item = &str> {
```
guilload, line 87: We almost always end up collecting into a set. Should we return a `HashSet<NodeId>` directly for `live_nodes`, `dead_nodes`, `seed_nodes`?

scuttlebutt/src/server.rs
```diff
@@ -172,30 +173,113 @@ impl UdpServer {
...
         // Update nodes liveliness
         let mut scuttlebutt_guard = self.scuttlebutt.lock().await;
         scuttlebutt_guard.update_nodes_liveliness();
     }
 
+    /// Selects a dead node to gossip with, with some probability.
+    fn select_dead_node_to_gossip_with(
+        &self,
+        rng: &mut SmallRng,
+        dead_nodes: &HashSet<String>,
```
guilload, line 251: Can we use `NodeId` whenever possible?

scuttlebutt/src/server.rs
```diff
@@ -172,30 +173,113 @@ impl UdpServer {
...
+            dead_nodes_count,
+        );
 
         // Drop lock to prevent deadlock in [`UdpSocket::gossip`].
         drop(scuttlebutt_guard);
+        let mut has_gossiped_with_a_seed_node = false;
         for node in &rand_nodes[..count] {
+            has_gossiped_with_a_seed_node =
+                has_gossiped_with_a_seed_node || seed_nodes.contains(node);
             let _ = self.gossip(node).await;
```
guilload, line 230: Can we log potential errors?

guilload (commented):
Since the complexity of the peer selection logic has increased, I'd isolate that logic from `gossip_multiple` so we can test it more easily.

scuttlebutt/src/server.rs
evanxg852000: The `cluster_nodes` will most likely contain seed nodes so I felt it's better. Besides this is just a case for the startup, in normal operation, we want live nodes.

scuttlebutt-test/Cargo.toml
```diff
@@ -21,4 +22,4 @@ tracing-subscriber = "0.3"
 assert_cmd = "2"
 predicates = "2"
 reqwest = { version = "0.11", default-features=false, features = ["blocking", "json"] }
-
+serial_test = "0.5.1"
```
guilload, line 25: Let's implement tests that can be run in parallel using random ports. Two tests can run parallel using random ports in different ranges.

scuttlebutt/src/server.rs
```diff
@@ -172,23 +174,42 @@ impl UdpServer {
...
-
-        // Select up to [`GOSSIP_COUNT`] node IDs at random.
-        let nodes = scuttlebutt_guard.cluster_state.nodes();
-        let count = nodes
-            .filter(|node_id| node_id != self_node_id)
-            .map(ToString::to_string)
-            .choose_multiple_fill(rng, &mut rand_nodes);
+        let cluster_state = scuttlebutt_guard.cluster_state();
+
+        let all_nodes = cluster_state
```
guilload, line 181:
```suggestion
        let peer_nodes = cluster_state
```

scuttlebutt/src/state.rs
```diff
@@ -90,12 +90,14 @@ impl NodeState {
 
 #[derive(Default, Serialize, Deserialize, Clone, Debug)]
 pub struct ClusterState {
+    seed_nodes: Vec<String>,
```
guilload, line 93:
```suggestion
    seed_nodes: HashSet<NodeId>,
```

guilload (approved):

scuttlebutt-test/Cargo.toml
fulmicoton: Not sure if this works here and universal this trick is, but when listenign to port 0, linux picks a random UDP port that is available.

Issue #6, opened 2022-01-08
Prevent KV with a value longer than 500B.


fulmicoton:
Right now scuttlebutt relies on UDP with a hardcoded mtu of 1KB.

evanxg852000:
Right now invariants like this are handled with `assert!` should we construct Error type and propagate or stick with `assert` for this case?

fulmicoton:
Good question. usually `Error` is better for an API like this, but I think we can stick to assert in this very case.

PR #18, opened 2022-02-14
Guard against large KV value.

evanxg852000:
Reject KV value with size more than 500B
[ Related Issue](https://github.com/quickwit-oss/scuttlebutt/issues/6)

### DIFF
--- a/scuttlebutt/src/state.rs
+++ b/scuttlebutt/src/state.rs
@@ -85,7 +85,14 @@ impl NodeState {
 
     fn set_with_version(&mut self, key: String, value: String, version: Version) {
         assert!(version > self.max_version);
-        assert!(value.bytes().len() <= MAX_KV_VALUE_SIZE, "Value too large.");
+        let value_size = value.bytes().len();
+        assert!(
+            value_size <= MAX_KV_VALUE_SIZE,
+            "Value for key `{}` is too large (actual: {}, maximum: {})",
+            key,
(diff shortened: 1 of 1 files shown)


scuttlebutt/src/state.rs
```diff
@@ -82,6 +85,7 @@ impl NodeState {
 
     fn set_with_version(&mut self, key: String, value: String, version: Version) {
         assert!(version > self.max_version);
+        assert!(value.bytes().len() <= MAX_KV_VALUE_SIZE, "Value too large.");
```
guilload, line 88: Is there an easy way to have a more user-friendly error message:
"Value for key `{}` is too large (actual: {}, maximum: {})`.

scuttlebutt/src/state.rs
```diff
@@ -320,6 +324,15 @@ mod tests {
         );
     }
 
+    #[test]
+    #[should_panic(expected = "Value too large.")]
+    fn test_cluster_state_set_with_large_value() {
+        let mut cluster_state = ClusterState::default();
+        let node_state = cluster_state.node_state_mut("node");
+        let large_value = "The quick brown fox jumps over the lazy dog.".repeat(12);
```
guilload, line 332: Cool, I didn't know that one :)

guilload (approved):

guilload (approved):

PR #22, opened 2022-03-03
NodeId implementation

evanxg852000:
Implements the concept of NodeId described [here](https://www.notion.so/Cluster-Node-Identification-83c2ba2568bc4a199aaf73d04bb98066) on the ScuttleButt .

### DIFF
--- a/scuttlebutt/src/lib.rs
+++ b/scuttlebutt/src/lib.rs
@@ -39,53 +39,76 @@ use tracing::{debug, error};
 use crate::digest::Digest;
 use crate::message::ScuttleButtMessage;
 use crate::serialize::Serializable;
-pub use crate::state::{ClusterState, SerializableClusterState};
 use crate::state::NodeState;
+pub use crate::state::{ClusterState, SerializableClusterState};
 
 /// Map key for the heartbeat node value.
 pub(crate) const HEARTBEAT_KEY: &str = "heartbeat";
 
 pub type Version = u64;
 
-// pub type NodeId = String;
-
+/// [`NodeId`] represents a ScuttleButt Node identifier.
+///
+/// For the lifetime of a cluster, nodes can go down and back up, they may
+/// permanently die. These are couple of issues we want to solve with [`NodeId`] struct:
+/// - We want a fresh local scuttlebutt state for every run of a node.
+/// - We don’t want other nodes to override a newly started node state with an obsolete state.
+/// - We want other running nodes to detect that a newly started node’s state prevails all its
+///   previous state.
+/// - We want a node to advertise its own gossip address.
+/// - We want a node to have an id that is the same across subsequent runs for keeping cache data
+///   around as long as possible.
+///
+/// Our solution to this is:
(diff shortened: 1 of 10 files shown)


scuttlebutt/src/lib.rs
```diff
@@ -39,15 +39,53 @@ use tracing::{debug, error};
...
 
 pub type Version = u64;
 
-pub type NodeId = String;
+// pub type NodeId = String;
+
+#[derive(Default, Clone, Hash, Eq, PartialEq, PartialOrd, Ord, Debug, Serialize, Deserialize)]
+pub struct NodeId {
+    id: String,
+    gossip_public_address: String,
```
fulmicoton, line 55: why not SocketAddr ?

scuttlebutt/src/lib.rs
```diff
@@ -39,15 +39,53 @@ use tracing::{debug, error};
...
 
 /// Map key for the heartbeat node value.
 pub(crate) const HEARTBEAT_KEY: &str = "heartbeat";
 
 pub type Version = u64;
 
-pub type NodeId = String;
+// pub type NodeId = String;
+
+#[derive(Default, Clone, Hash, Eq, PartialEq, PartialOrd, Ord, Debug, Serialize, Deserialize)]
```
fulmicoton, line 52: We need doc here. Copy pasting your doc is ok.

scuttlebutt/src/lib.rs
```diff
@@ -39,15 +39,53 @@ use tracing::{debug, error};
...
+pub struct NodeId {
+    id: String,
+    gossip_public_address: String,
+}
+
+impl NodeId {
+    fn new(id: &str, gossip_public_address: &str) -> Self {
```
fulmicoton, line 59: That should be `String`, or T: ToString if it is too painful in tests.

scuttlebutt/src/lib.rs
fulmicoton, line 55: why did you remove `node_generation`

evanxg852000: hmm, I needed to implement Default, not sure of a good default SocketAddr and this is not an option really.  Also, this made testing easier because I could `NodeId::from("node-one")` without worrying of passing a valid SocketAdrr.

scuttlebutt/src/lib.rs
evanxg852000: Adding doc, also part of the doc will be in Quickwit where QuickwitNodeId is defined.

scuttlebutt/src/lib.rs
evanxg852000: It will be available from the client-side (Quickwit). See the docs
while writing the docs, I noticed we can just use `{node_unique_id}/{node_generation}` instead of `{node_unique_id}/{gossip_public_address}/{generation}`
https://github.com/quickwit-oss/scuttlebutt/pull/22/files#diff-9a1c8c0b2fe22d61843aec8d062321324f330717737a4880c634fc3c5a7b43fbR74

fulmicoton (approved):

PR #29, opened 2022-03-14
Add license

evanxg852000:
Set license field to `AGPL-3.0-or-later` because we already have a license header describing AGPL

### DIFF
--- a/scuttlebutt/Cargo.toml
+++ b/scuttlebutt/Cargo.toml
@@ -2,6 +2,10 @@
 name = "scuttlebutt"
 version = "0.1.0"
 edition = "2021"
+license = "AGPL-3.0-or-later"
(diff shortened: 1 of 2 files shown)


scuttlebutt/Cargo.toml
```diff
@@ -2,6 +2,10 @@
 name = "scuttlebutt"
 version = "0.1.0"
 edition = "2021"
+license = "AGPL-3.0-or-later"
+authors = ["Quickwit, Inc. <hello@quickwit.io>"]
+description = "Cluster membershipe management library implemented using Scuttlebutt."
```
fulmicoton, line 7:
```suggestion
description = "Cluster membership management library implemented using Scuttlebutt."
```

Issue #17, opened 2022-02-10
Remove dead node from the cluster node list eventually 


evanxg852000:
Currently, when a cluster node dies, It is kept in the cluster node even if the node never comes back.
This can make the cluster node list grow very large for a long-running node. 
We want to remove a dead node from the cluster node list at some point.

PR #33, opened 2022-03-28
Node garbage collection

evanxg852000:
Implemented dead nodes garbage collection.

https://github.com/quickwit-oss/scuttlebutt/issues/17

### DIFF
--- a/scuttlebutt/src/failure_detector.rs
+++ b/scuttlebutt/src/failure_detector.rs
@@ -39,7 +39,7 @@ pub struct FailureDetector {
     /// Denotes live nodes.
     live_nodes: HashSet<NodeId>,
     /// Denotes dead nodes.
-    dead_nodes: HashSet<NodeId>,
+    dead_nodes: HashMap<NodeId, Instant>,
 }
 
 impl FailureDetector {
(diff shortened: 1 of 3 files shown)


scuttlebutt/src/state.rs
```diff
@@ -130,6 +130,10 @@ impl ClusterState {
         self.seed_nodes.iter().map(|node_id| node_id.as_str())
     }
 
+    pub fn remove_node(&mut self, node_id: &NodeId) {
```
fulmicoton, line 133: should that be pub?

scuttlebutt/src/failure_detector.rs
```diff
@@ -72,7 +72,7 @@ impl FailureDetector {
             debug!(node_id = ?node_id, phi = phi, "updating node liveliness");
             if phi > self.config.phi_threshold {
                 self.live_nodes.remove(node_id);
```
fulmicoton, line 74: super ultra nitpick.

I like the idea to emphasive on the fact taht we move the node from live to dead.

for instance via:

```
if let Some(node_id) = self.live_nodes.remove(node_id) {
    self.dead_nodes.insert(node_id, Instant::now());
}
```

fulmicoton: ( no need to fix that'ss probably just me)

fulmicoton (approved):

scuttlebutt/src/failure_detector.rs
evanxg852000: This might not be a nitpick. it really makes the code speaks for itself TBH. which is always great to have

PR #36, opened 2022-04-07
Add cluster_name property to avoid several clusters merging accidentally

xvello:
As described in https://github.com/quickwit-oss/quickwit/pull/1260, we need a mechanism to ensure clusters do not accidentally merge when IPs are reused.

This PR:

- introduces a required `cluster_name` argument when setting up the cluster
- adds that cluster name to outgoing `Syn` messages
- adds a `BadCluster` message type
- if the cluster name in the `Syn` message does not match the peer's cluster name, a `BadCluster` message is returned instead of a `SynAck`

## Additional notes

- This PR changes the wire format, but I did not see any versioning mechanism in the protocol. We should merge this before v0.3 is released, to sidestep any compatibility issue.

### DIFF
--- a/scuttlebutt/src/lib.rs
+++ b/scuttlebutt/src/lib.rs
@@ -186,7 +186,7 @@ impl ScuttleButt {
             } => {
                 if cluster_name != self.cluster_name {
                     warn!(
-                        cluster_name = ?cluster_name,
+                        cluster_name = %cluster_name,
                         "rejecting syn message with mismatching cluster name"
                     );
                     return Some(ScuttleButtMessage::BadCluster);
--- a/scuttlebutt/src/lib.rs
+++ b/scuttlebutt/src/lib.rs
@@ -187,9 +187,9 @@ impl ScuttleButt {
                 if cluster_name != self.cluster_name {
                     warn!(
                         cluster_name = ?cluster_name,
-                        "ignoring syn message with mismatching cluster name"
+                        "rejecting syn message with mismatching cluster name"
                     );
-                    return None;
+                    return Some(ScuttleButtMessage::BadCluster);
                 }
 
                 let self_digest = self.compute_digest();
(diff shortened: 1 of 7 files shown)


fulmicoton (changes requested):
As you suggested in the PR description, could you add an error message to the protocol?

As you mentioned it is nicer for testing, but also,I suspect a common problem that will occur is a misconfigured typo in the conf of one of the server. The user will likely investigate the log of the server that cannot connect to the cluster first and an explicit error message there might be nice.

fulmicoton:
I create #37 and #38 too.

xvello:
I just pushed a commit introducing a `BadCluster` message, instead of silently dropping the message.

I started writing the code with a more generic `Rst` message, but I could not get a relevant error displayed on the misconfigured node without extending it with a `RstReason` enum, which added unneeded complexity.

xvello:
`ignore_oversized_payload` is failing, but I think this test has lost its initial meaning and is not useful anymore. Discussing it on discord.

scuttlebutt/src/lib.rs
```diff
@@ -169,12 +172,26 @@ impl ScuttleButt {
...
     pub fn process_message(&mut self, msg: ScuttleButtMessage) -> Option<ScuttleButtMessage> {
         match msg {
-            ScuttleButtMessage::Syn { digest } => {
+            ScuttleButtMessage::Syn {
+                cluster_name,
+                digest,
+            } => {
+                if cluster_name != self.cluster_name {
+                    warn!(
+                        cluster_name = ?cluster_name,
```
fulmicoton, line 189: %

fulmicoton, line 189:
```suggestion
                        cluster_name = %cluster_name,
```

fulmicoton (approved):

fulmicoton:
@xvello FYI: in tracing ?val will use Debug. %val will call Display.
For a String you want %val 

Alternatively, you can also try to get a &str and avoid any modifier. `key=val.as_str()`

PR #46, opened 2022-04-30
Add a simple example of setting a key at runtime.

canardleteer:
I added an additional endpoint to `chitchat-test` to show how to set a nodes key's at runtime.

### DIFF
--- a/chitchat-test/src/lib.rs
+++ b/chitchat-test/src/lib.rs
@@ -8,3 +8,8 @@ pub struct ApiResponse {
     pub live_nodes: Vec<NodeId>,
     pub dead_nodes: Vec<NodeId>,
 }
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct SetKeyValueResponse {
+    pub status: bool,
+}
--- a/chitchat-test/src/main.rs
+++ b/chitchat-test/src/main.rs
@@ -32,7 +32,7 @@ impl Api {
 
     /// Set a key & value on this node (with no validation).
     #[oai(path = "/set_kv/", method = "get")]
-    async fn named_visitor(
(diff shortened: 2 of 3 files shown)


chitchat-test/src/lib.rs
```diff
@@ -8,3 +8,8 @@ pub struct ApiResponse {
     pub live_nodes: Vec<NodeId>,
     pub dead_nodes: Vec<NodeId>,
 }
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct SetKeyValueResponse {
```
fulmicoton, line 13: that should be in the main file

fulmicoton (approved):
Can we add a unit test for it?

chitchat-test/src/main.rs
```diff
@@ -28,6 +29,21 @@ impl Api {
         };
         Json(serde_json::to_value(&response).unwrap())
     }
+
+    /// Set a key & value on this node (with no validation).
+    #[oai(path = "/set_kv/", method = "get")]
+    async fn named_visitor(
```
fulmicoton, line 35:
```suggestion
    async fn set_kv(
```

chitchat-test/src/main.rs
```diff
@@ -28,6 +29,21 @@ impl Api {
         };
         Json(serde_json::to_value(&response).unwrap())
     }
+
+    /// Set a key & value on this node (with no validation).
+    #[oai(path = "/set_kv/", method = "get")]
+    async fn set_kv(
```
fulmicoton, line 35:
```suggestion
    async fn set_kv_handler(
```

fulmicoton (approved):

PR #47, opened 2022-05-06
Update run-servers.sh

sysarcher:
Use the `-h` command line arg instead of `--host`

Error was:

```
error: Found argument '--host' which wasn't expected, or isn't valid in this context
```


evanxg852000:
I think it's better we fix it by adding the long form of the argument `host`.

fulmicoton:
Closing, we swithced to `--listen_addr`

sysarcher:
Awesome... sorry I couldn't reply earlier.. neck deep in work. 

Thanks for the project though.

PR #48, opened 2022-05-12
Various refactoring and bug fix

fulmicoton:
# Removed the use of string everywhere for addresses

The string were hiding several bugs.
This PR partially addresses it.

Before this PR, all addresses were `String`, that were
eventually resolved via `ToSocketAddrs`.

There are several cons and one bug associated with this approach.
Readability was hurt and API contract enforcing was a bit broken.
For instance, a user could have supplied an unresolved hostname as a gossip address.
As a result, all calls to that node would have triggered a DNS request. 

In addition, the logic that checks for peers to gossip with takes contains
some logic that relies on equality. For instance, we check if no seed was
already picked, before artificially adding a seed candidate.

That logic is broken, because the seed addrs are likely to be unresolved names
while peer address are likely to be socket addresses.

We switched all address to socket address except seeds.
Supplying a hostname as a socket addr is something that is very useful for k8s
for instance.

Internally, we just do slow polling over the name resolution of this seed hostname
to detect eventual new seed servers.

# Added node_id, public_addr option in chitchat test program

# simplified subprocess kill logic in unit test

Rather than registering a panic handler, and using a global, we rely on Drop.

# Seeding the random generator with the thread local generator (itself has been seed using system entropy).
This might help speed up the unit test a little bit.

### DIFF
--- a/chitchat/src/configuration.rs
+++ b/chitchat/src/configuration.rs
@@ -1,0 +1,47 @@
+use std::net::SocketAddr;
+use std::time::Duration;
+
+use crate::{FailureDetectorConfig, NodeId};
+
+pub struct ChitchatConfig {
+    pub node_id: NodeId,
+    pub cluster_id: String,
+    pub gossip_interval: Duration,
+    pub listen_addr: SocketAddr,
+    pub seed_nodes: Vec<String>,
+    pub mtu: usize,
+    pub failure_detector_config: FailureDetectorConfig,
+}
+
+impl ChitchatConfig {
+    #[cfg(test)]
+    pub fn for_test(port: u16) -> Self {
+        let node_id = NodeId::for_test_localhost(port);
+        let listen_addr = node_id.gossip_public_address;
+        Self {
+            node_id,
+            cluster_id: "default-cluster".to_string(),
+            gossip_interval: Duration::from_millis(50),
+            listen_addr,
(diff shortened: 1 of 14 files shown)


chitchat/src/configuration.rs
```diff
@@ -0,0 +1,47 @@
+use std::net::SocketAddr;
+use std::time::Duration;
+
+use crate::{FailureDetectorConfig, NodeId};
+
+pub struct ChitchatConfig {
```
evanxg852000, line 6:
```suggestion
/// A struct for configuring a Chitact instance.
pub struct ChitchatConfig {
```

chitchat/src/server.rs
```diff
@@ -4,75 +4,150 @@ use std::sync::Arc;
...
+// The latter is especially important when relying on
+// a headless service in k8s or when using DNS in general.
+//
+// In that case, we do not want to perform the resolution
+// once and forall.
+// We want to periodically retry DNS resolution,
+// in order to avoid having a split cluster.
+//
+// The newcomers are supposed to chime in too,
+// so there is not need to refresh it too often,
```
evanxg852000, line 81:
```suggestion
// so there is no need to refresh it too often,
```

evanxg852000 (approved):

PR #56, opened 2022-05-16
Adding a transport abstraction

fulmicoton:
This PR is adding a transport abstraction.

The point of this transport layer is:
-  to make it easy to simulate what happens if X% of the message are dropped, or if they are delayed.
- avoid having to juggle with UDP port allocation in unit test, and safely simulate 100 of nodes. 
- measure metrics such as the amount of data exchanged during simulations.
- slightly simplify the code of the server. 

This PR also:
Stops updating heartbeat on all ack messages.
While this was not hurting correctness, in the same round of gossip, we would send 3 different gossip to peers A,B,C. 
This would generate meaningless update messages within the cluster.

This PR also adds a lot of performance tests simulating chitchat and checking that it is behaving correctly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,4 +40,4 @@ jobs:
         name: cargo test
         with:
           command: test
-          args: -- --test-threads 1
+          args: --release -- --test-threads 1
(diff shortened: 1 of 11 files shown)


evanxg852000 (approved):
Great abstraction on the transport layer :+1:

PR #66, opened 2022-08-23
Make it possible to set a ready predicate for nodes.

fulmicoton:


chitchat/src/configuration.rs
```diff
@@ -11,6 +14,11 @@ pub struct ChitchatConfig {
     pub listen_addr: SocketAddr,
     pub seed_nodes: Vec<String>,
     pub failure_detector_config: FailureDetectorConfig,
+    // `is_ready_predicate` makes it possible for node to advertise themselves as not "ready",
+    // for instance if they are `starting` or if they lost connection to a third-part service.
```
fmassot, line 18:
```suggestion
    // `is_ready_predicate` makes it possible for a node to advertise itself as not "ready".
    // For instance, if it is `starting` or if it lost connection to a third-party service.
```

chitchat/src/lib.rs
```diff
@@ -264,6 +268,23 @@ impl Chitchat {
         self.failure_detector.live_nodes()
     }
 
+    /// Retrieves the list of nodes that are ready.
+    /// To be ready, a node has to be live and pass the `is_ready_filter` as
```
fmassot, line 272:
```suggestion
    /// To be ready, a node has to be alive and pass the `is_ready_filter` as
```

fmassot (approved):

PR #77, opened 2023-02-08
Add deletion key marker and garbage collection.

fmassot:
TODO:

- [ ] Add test on `compute_delta` to check a new node is not reset.

### DIFF
--- a/chitchat/src/configuration.rs
+++ b/chitchat/src/configuration.rs
@@ -19,6 +19,16 @@ pub struct ChitchatConfig {
     //
     // If `None`, a node is ready as long as it is alive.
     pub is_ready_predicate: Option<Box<dyn Fn(&NodeState) -> bool + Send>>,
+    // Marked for deletion grace period expressed as a number of version threshold.
+    // Chitchat ensures a marked for deletion key is eventually deleted by three mecanisms:
+    // - Garbage collection: each heartbeat, marked for deletion keys with `key_version +
+    //   marked_for_deletion_grace_period < node.max_version` are deleted.
+    // - Compute delta: for a given node, if `digest_node_max_version +
+    //   marked_for_deletion_grace_period < node_max_version`, the node is flagged "to be reset"
+    //   and the delta is populated with all keys and values.
+    // - Apply delta: for a node flagged "to be reset", Chitchat will remove the node state and
+    //   populate a fresh new node state with the keys and values present in the delta.
+    pub marked_for_deletion_grace_period: usize,
 }
 
 impl ChitchatConfig {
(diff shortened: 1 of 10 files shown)


chitchat/src/configuration.rs
```diff
@@ -19,6 +19,16 @@ pub struct ChitchatConfig {
...
     pub is_ready_predicate: Option<Box<dyn Fn(&NodeState) -> bool + Send>>,
+    // Marked for deletion grace period expressed as a number of version threshold.
+    // Chitchat ensures a marked for deletion key is eventually deleted by three mecanisms:
+    // - Garbage collection: each heartbeat, marked for deletion keys with `key_version +
+    //   marked_for_deletion_grace_period < node.max_version` are deleted.
+    // - Compute delta: for a given node, if `digest_node_max_version +
+    //   marked_for_deletion_grace_period < node_max_version`, the node is flagged "to be reset"
+    //   and the delta is populated with all keys and values.
+    // - Apply delta: for a node flagged "to be reset", Chitchat will remove the node state and
+    //   populate a fresh new node state with the keys and values present in the delta.
```
fulmicoton, line 30: briliant comment.

chitchat/src/delta.rs
```diff
@@ -16,6 +17,10 @@ impl Serializable for Delta {
             node_id.serialize(buf);
             node_delta.serialize(buf);
         }
+        (self.nodes_to_reset.len() as u16).serialize(buf);
```
fulmicoton, line 20:
```suggestion
       u16::try_from(self.nodes_to_reset.len()).unwrap().serialize(buf);
```

would be a little bit more defensive.

chitchat/src/state.rs
```diff
@@ -32,16 +32,24 @@ impl Default for NodeState {
...
     ) -> impl Iterator<Item = (&str, &VersionedValue)> {
         // TODO optimize by checking the max version.
-        self.key_values
-            .iter()
-            .filter(move |&(_key, versioned_value)| versioned_value.version > floor_version)
-            .map(|(key, record)| (key.as_str(), record))
+        self.iter_key_values(move |_key, versioned_value| versioned_value.version > floor_version)
     }
 
     pub fn get(&self, key: &str) -> Option<&str> {
```
fulmicoton, line 55: Don't we want to hide values that are `marked_for_deletion`?

chitchat/src/state.rs
```diff
@@ -32,16 +32,24 @@ impl Default for NodeState {
 }
 
 impl NodeState {
+    /// Returns an iterator over keys matching the given predicate.
+    pub fn iter_key_values(
+        &self,
+        predicate: impl Fn(&String, &VersionedValue) -> bool,
+    ) -> impl Iterator<Item = (&str, &VersionedValue)> {
```
fulmicoton, line 39: If this is public API think we should hide deleted entries.

fulmicoton (approved):

PR #80, opened 2023-04-11
Refactor `NodeId`

guilload:
Chitchat and Quickwit use `NodeId` for different things, which makes the code in Quickwit very confusing.  So, I choose to rename `NodeId` to `ChitchatId`.

In addition, the Chitchat doc explains why you probably want to use a secondary id to detect node restarts, but then forces you into dealing with it yourself by appending some numeric ID to the node ID, which is cumbersome. In addition, encoding this ID as an integer actually consumes less bytes (8) than the string suffix alternative (13 bytes for a timestamp in millis + 1 for the separator char).

### DIFF
--- a/chitchat/src/lib.rs
+++ b/chitchat/src/lib.rs
@@ -47,71 +47,52 @@ const MAX_UDP_DATAGRAM_PAYLOAD_SIZE: usize = 65_507;
 
 pub type Version = u64;
 
-/// [`NodeId`] represents a Chitchat Node identifier.
+/// For the lifetime of a cluster, nodes can go down and come back up multiple times. They may also
+/// die permanently. A [`ChitchatId`] is composed of three components:
+/// - `node_id`: an identifier unique across the cluster.
+/// - `generation_id`: a numeric identifier that distinguishes a node's states between restarts.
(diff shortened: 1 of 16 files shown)


chitchat/src/lib.rs
```diff
@@ -47,71 +47,52 @@ const MAX_UDP_DATAGRAM_PAYLOAD_SIZE: usize = 65_507;
 
 pub type Version = u64;
 
-/// [`NodeId`] represents a Chitchat Node identifier.
+/// For the lifetime of a cluster, nodes can go down and come back up multiple times. They may also
```
fmassot, line 50: Really nice docs, it's shorter and more accurate, I like to be explicit with the `join_id`.

Nitpick: you can ignore but I have a small preference for generation ID, how did you come up with `join_id` (google? :))?

fmassot (approved):

chitchat/src/lib.rs
guilload: Generation ID is good too. I'll go for that.

guilload:
> `assert_cmd v2.0.11` cannot be built because it requires rustc 1.65.0 or newer, while the currently active rustc version is 1.64.0

t_t

PR #82, opened 2023-04-18
Move heartbeat to dedicated `NodeState` attribute instead of generic key

guilload:


chitchat/src/state.rs
```diff
@@ -1,21 +1,23 @@
...
-use crate::{ChitchatId, Version, VersionedValue};
+use crate::digest::{Digest, NodeDigest};
+use crate::{ChitchatId, Heartbeat, MaxVersion, Version, VersionedValue};
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct NodeState {
+    pub(crate) heartbeat: Heartbeat,
     pub(crate) key_values: BTreeMap<String, VersionedValue>,
-    pub(crate) max_version: u64,
+    pub(crate) max_version: MaxVersion,
```
fmassot, line 20: Remark that you can ignore, I think this is a discussion we can have if we want to write some coding style guidelines.

I'm not convinced by using a type alias to replace a simple type such as u64. I don't see the added value and generally, I found the code easier to understand when I see the real type `u64`.

chitchat/src/state.rs
```diff
@@ -50,11 +61,11 @@ impl NodeState {
     ) -> impl Iterator<Item = (&str, &VersionedValue)> {
         self.key_values
             .iter()
-            .filter(move |&(key, versioned_value)| predicate(key, versioned_value))
+            .filter(move |(key, versioned_value)| predicate(key, versioned_value))
             .map(|(key, record)| (key.as_str(), record))
     }
 
-    /// Returns an iterator over the version values that are older than `floor_version`.
+    /// Returns an iterator over the version values that are strictly older than `floor_version`.
```
fmassot, line 68: I would use `greater` instead of `old`. Greater version are in fact more "recent".

chitchat/src/state.rs
```diff
@@ -165,43 +187,43 @@ impl ClusterState {
...
         for (chitchat_id, node_delta) in delta.node_deltas {
-            let mut node_state_map = self
-                .node_states
-                .entry(chitchat_id)
-                .or_insert_with(NodeState::default);
+            let mut node_state = self.node_states.entry(chitchat_id).or_default();
+
+            if node_state.heartbeat < node_delta.heartbeat {
+                node_state.heartbeat = node_delta.heartbeat;
+                node_state.last_heartbeat = Instant::now();
```
fmassot, line 197: I'm wondering why we need `last_heartbeat`. When looking at the code, it seems useless.

chitchat/src/state.rs
```diff
@@ -165,43 +187,43 @@ impl ClusterState {
...
+                let entry = node_state.key_values.entry(key);
                 match entry {
-                    Entry::Occupied(mut record) => {
-                        if record.get().version >= versioned_value.version {
-                            // Due to the message passing being totally asynchronous, it is not an
-                            // error to receive updates that are already obsolete.
-                            continue;
+                    Entry::Occupied(mut occupied_entry) => {
+                        let local_versioned_value = occupied_entry.get_mut();
+                        // Due to the message passing being totally asynchronous, it is not an
```
fmassot, line 206: The comment is a bit weird now as you remove the "continue" :)
I would change it slightly and say something like this

```
Obsolete values are ignored. Due to the message passing being totally asynchronous, it is not a...
```

chitchat/src/server.rs
```diff
@@ -37,11 +37,13 @@ async fn dns_refresh_loop(
     seed_addrs_tx: watch::Sender<HashSet<SocketAddr>>,
 ) {
     let mut interval = time::interval(DNS_POLLING_DURATION);
-    // We actually do not want to run the polling loop right away,
+    // We actually DO want to run the polling loop right away,
```
fmassot, line 40: Ok the comment sounds strange :)

We do want the polling but we made the first one already before.

The comment could be 

```
The caller has just run the dns resolution. We can wait one tick before retrying.   
```

fmassot: Also, now there are two ticks. Is this what you wanted?

PR #98, opened 2023-11-16
Added API to listen to event start by a given prefix.

fulmicoton:


chitchat/src/listener.rs
```diff
@@ -0,0 +1,246 @@
+use std::collections::{BTreeMap, HashMap};
+use std::ops::Bound;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::{Arc, RwLock, Weak};
+
+pub struct ListenerHandle {
+    prefix: String,
+    idx: usize,
```
guilload, line 8: I haven't read the rest of the file yet, so I have no idea what that means. Is that the handle ID or some index into something else? A better name/comment would help.

chitchat/src/listener.rs
```diff
@@ -0,0 +1,246 @@
...
+        &self,
+        key_prefix: String,
+        boxed_listener: BoxedListener,
+    ) -> ListenerHandle {
+        let key_prefix = key_prefix.to_string();
+        let weak_listeners = Arc::downgrade(&self.inner);
+        let mut inner_listener_guard = self.inner.write().unwrap();
+        let new_idx = inner_listener_guard
+            .listener_idx
+            .fetch_add(1, Ordering::SeqCst);
```
guilload, line 60: If you're never reading the atomic variable, `Ordering::Relaxed` will do just fine.

guilload (approved):
Turns out it was a `listener_id` :)

chitchat/src/listener.rs
fulmicoton: yes for an id we can do with relaxed.

PR #101, opened 2023-11-29
Added node id to the change event

fulmicoton:


chitchat/src/lib.rs
```diff
@@ -282,7 +282,8 @@ impl Chitchat {
         self.cluster_state.compute_digest(dead_nodes)
     }
 
-    /// Subscribe a callback that will be called every time a key matching the  supplied prefix.
+    /// Subscribe a callback that will be called every time a key matching the supplied prefix
+    /// is inserted or is updated.
```
guilload, line 286:
```suggestion
    /// Subscribes a callback that will be called every time a key matching the supplied prefix
    /// is inserted or updated.
```

chitchat/src/lib.rs
```diff
@@ -291,20 +292,40 @@ impl Chitchat {
...
 
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub struct KeyChangeEvent<'a> {
+    key: &'a str,
+    value: &'a str,
+    node: &'a ChitchatId,
+}
+
+impl<'a> KeyChangeEvent<'a> {
+    pub fn strip_key_prefix(&self, prefix: &str) -> Option<KeyChangeEvent> {
```
guilload, line 319: Does it need to be public?

guilload (approved):

chitchat/src/lib.rs
fulmicoton: no. removing pub

chitchat/src/lib.rs
fulmicoton: manually edited the text.

PR #106, opened 2024-01-12
Make serialize module public to align with transport module

melbourne2991:
Addresses https://github.com/quickwit-oss/chitchat/issues/78


fulmicoton:
Closing as Serializable was reexported as a pub.

PR #108, opened 2024-01-12
forget previous generation

### LINKED ISSUE #94: Generation increase in chitchat does not really clean up the state
To reproduce.

Create two nodes `node_{1,2}`.
Restart node two.

You will end up with a two nodes 2 in the state.Both staying in the READY mode, despite the heartbeat never being updated.

```json
{
  "cluster_id": "quickwit-default-cluster",
  "self_node_id": "node-1",
  "ready_nodes": [
    {
      "node_id": "node-1",
      "generation_id": 1699243351019449000,
      "gossip_advertise_addr": "127.0.0.1:7280"
    },
    {
      "node_id": "node-2",
      "generation_id": 1699243416590203000,
      "gossip_advertise_addr": "127.0.0.1:8280"
    }
  ],
  "live_nodes": [],
  "dead_nodes": [
    {
      "node_id": "node-2",
      "generation_id": 1699242914104763000,
      "gossip_advertise_addr": "127.0.0.1:8280"
    }
  ],
  "chitchat_state_snapshot": {
    "node_state_snapshots": [
      {
        "chitchat_id": {
          "node_id": "node-1",
          "generation_id": 1699243351019449000,
          "gossip_advertise_addr": "127.0.0.1:7280"
        },
        "node_state": {
          "heartbeat": 487,
          "key_values": {
            "enabled_services": {
              "value": "janitor,control_plane,indexer,searcher,metastore",
              "version": 1,
              "tombstone": null
            },
            "grpc_advertise_addr": {
              "value": "127.0.0.1:7281",
              "version": 2,
              "tombstone": null
            },
            "indexer_capacity": {
              "value": "2000",
              "version": 4,
              "tombstone": null
            },
            "indexing_task:cflogs:01HE82PCCWZNHSZKVZSJ2ARB73:_ingest-api-source": {
              "value": "1",
              "version": 117,
              "tombstone": null
            },
            "indexing_task:otel-logs-v0_6:01HE82P019G3THGA3JK9Q833WA:_ingest-api-source": {
              "value": "1",
              "version": 118,
              "tombstone": null
            },
            "indexing_task:otel-traces-v0_6:01HE82P02NKPZX17J3304XY4T2:_ingest-api-source": {
              "value": "1",
              "version": 119,
              "tombstone": null
            },
            "readiness": {
              "value": "READY",
              "version": 116,
              "tombstone": null
            }
          },
          "max_version": 119
        }
      },
      {
        "chitchat_id": {
          "node_id": "node-2",
          "generation_id": 1699242903139100000,
          "gossip_advertise_addr": "127.0.0.1:8280"
        },
        "node_state": {
          "heartbeat": 3,
          "key_values": {
            "enabled_services": {
              "value": "indexer,searcher",
              "version": 1,
              "tombstone": null
            },
            "grpc_advertise_addr": {
              "value": "127.0.0.1:8281",
              "version": 2,
              "tombstone": null
            },
            "indexer_capacity": {
              "value": "4000",
              "version": 4,
              "tombstone": null
            },
            "readiness": {
              "value": "NOT_READY",
              "version": 3,
              "tombstone": null
            }
          },
          "max_version": 4
        }
      },
      {
        "chitchat_id": {
          "node_id": "node-2",
          "generation_id": 1699242914104763000,
          "gossip_advertise_addr": "127.0.0.1:8280"
        },
        "node_state": {
          "heartbeat": 492,
          "key_values": {
            "enabled_services": {
              "value": "indexer,searcher,control_plane",
              "version": 1,
              "tombstone": null
            },
            "grpc_advertise_addr": {
              "value": "127.0.0.1:8281",
              "version": 2,
              "tombstone": null
            },
            "indexer_capacity": {
              "value": "4000",
              "version": 4,
              "tombstone": null
            },
            "indexing_task:cflogs:01HE82PCCWZNHSZKVZSJ2ARB73:_ingest-api-source": {
              "value": "1",
              "version": 114,
              "tombstone": null
            },
            "indexing_task:otel-logs-v0_6:01HE82P019G3THGA3JK9Q833WA:_ingest-api-source": {
              "value": "1",
              "version": 115,
              "tombstone": null
            },
            "indexing_task:otel-traces-v0_6:01HE82P02NKPZX17J3304XY4T2:_ingest-api-source": {
              "value": "1",
              "version": 116,
              "tombstone": null
            },
            "readiness": {
              "value": "READY",
              "version": 113,
              "tombstone": null
            }
          },
          "max_version": 116
        }
      },
      {
        "chitchat_id": {
          "node_id": "node-2",
          "generation_id": 1699243416590203000,
          "gossip_advertise_addr": "127.0.0.1:8280"
        },
        "node_state": {
          "heartbeat": 421,
          "key_values": {
            "enabled_services": {
              "value": "control_plane,searcher",
              "version": 1,
              "tombstone": null
            },
            "grpc_advertise_addr": {
              "value": "127.0.0.1:8281",
              "version": 2,
              "tombstone": null
            },
            "readiness": {
              "value": "READY",
              "version": 45,
              "tombstone": null
            }
          },
          "max_version": 45
        }
      }
    ],
    "seed_addrs": []
  }
}
```

trinity-1686a:
fix #94 

I figured what was preventing gc.
- Node1 and Node2 communicate.
- Node2 dies.
- Node3 appears.
- Node1 isn't aware yet that Node2 is dead; it gives info about Node2 to Node3
- Node3 receives info of Node2, it record a heartbeat, but Node2 isn't in the state, so the failure detector actually doesn't record anything. Then the state is updated.
- Node1 and Node3 don't communicate about Node2 as nothing changed
- Node1 discovers Node2 is dead
- after some time, Node1 finally forget Node2
- Node3 notice Node1 haven't heard of Node2, and sends it the info
- Node1 receives info of Node2, it record a heartbeat, but Node2 isn't in the state, so the failure detector actually doesn't record anything. Then the state is updated.

Both Node1 and Node3 knows about Node2, but none track its state in their failure detector

what I implemented to fix that is that on first learning about a node, it's recorded in the failure detector as dead. if we get a 2nd heartbeat (with higher sequence number), we update to alive. That way a node in state should always be in the failure detector. We add it as dead because otherwise each time Node1 forget, the message from Node3 makes it believe that Node2 is alive, then Node3 forget, Node1 makes it believe Node2 is alive, then Node1 forget.... (this is only an issue if `dead_node_grace_period` can be smaller than the time it takes to detect a node is dead)

<br>
  
currently tests don't pass, and I have not been successful at fixing them. I left a comment with what I've figured so far at the place where it fails. Any help welcome!

### DIFF
--- a/chitchat/src/failure_detector.rs
+++ b/chitchat/src/failure_detector.rs
@@ -48,6 +48,19 @@ impl FailureDetector {
         heartbeat_window.report_heartbeat();
     }
 
+    pub fn report_unknown(&mut self, chitchat_id: &ChitchatId) {
+        debug!(node_id=%chitchat_id.node_id, "reporting unknown node heartbeat.");
+        self.node_samples
+            .entry(chitchat_id.clone())
+            .or_insert_with(|| {
+                SamplingWindow::new(
+                    self.config.sampling_window_size,
+                    self.config.max_interval,
+                    self.config.initial_interval,
+                )
+            });
+    }
+
     /// Marks the node as dead or alive based on the current phi value.
     pub fn update_node_liveness(&mut self, chitchat_id: &ChitchatId) {
         if let Some(phi) = self.phi(chitchat_id) {
@@ -181,10 +194,14 @@ impl SamplingWindow {
 
     /// Computes the sampling window's phi value.
     pub fn phi(&self) -> f64 {
-        // Ensure we don't call before any sample arrival.
-        assert!(self.intervals.mean() > 0.0 && self.last_heartbeat.is_some());
-        let elapsed_time = self.last_heartbeat.unwrap().elapsed().as_secs_f64();
-        elapsed_time / self.intervals.mean()
+        if self.last_heartbeat.is_none() {
+            // if we phi is called before we have a sample, we assume the node isn't really alive.
+            f64::INFINITY
+        } else {
+            assert!(self.intervals.mean() > 0.0);
+            let elapsed_time = self.last_heartbeat.unwrap().elapsed().as_secs_f64();
+            elapsed_time / self.intervals.mean()
+        }
     }
 }
 
--- a/chitchat/src/failure_detector.rs
+++ b/chitchat/src/failure_detector.rs
@@ -194,13 +194,13 @@ impl SamplingWindow {
 
     /// Computes the sampling window's phi value.
     pub fn phi(&self) -> f64 {
-        if self.last_heartbeat.is_none() {
+        if let Some(last_heartbeat) = self.last_heartbeat {
+            assert!(self.intervals.mean() > 0.0);
+            let elapsed_time = last_heartbeat.elapsed().as_secs_f64();
+            elapsed_time / self.intervals.mean()
+        } else {
             // if we phi is called before we have a sample, we assume the node isn't really alive.
             f64::INFINITY
-        } else {
-            assert!(self.intervals.mean() > 0.0);
-            let elapsed_time = self.last_heartbeat.unwrap().elapsed().as_secs_f64();
-            elapsed_time / self.intervals.mean()
         }
     }
 }
--- a/chitchat/src/failure_detector.rs
+++ b/chitchat/src/failure_detector.rs
@@ -8,18 +8,18 @@ use mock_instant::Instant;
 use serde::{Deserialize, Serialize};
 use tracing::debug;
 
-use crate::{ChitchatId, ChitchatIdNodeEq};
+use crate::ChitchatId;
 
 /// A phi accrual failure detector implementation.
 pub struct FailureDetector {
     /// Heartbeat samples for each node.
-    node_samples: HashMap<ChitchatIdNodeEq, SamplingWindow>,
+    node_samples: HashMap<ChitchatId, SamplingWindow>,
     /// Failure detector configuration.
     config: FailureDetectorConfig,
     /// Denotes live nodes.
-    live_nodes: HashSet<ChitchatIdNodeEq>,
+    live_nodes: HashSet<ChitchatId>,
     /// Denotes dead nodes.
-    dead_nodes: HashMap<ChitchatIdNodeEq, Instant>,
+    dead_nodes: HashMap<ChitchatId, Instant>,
 }
 
 impl FailureDetector {
@@ -37,7 +37,7 @@ impl FailureDetector {
         debug!(node_id=%chitchat_id.node_id, "reporting node heartbeat.");
         let heartbeat_window = self
             .node_samples
-            .entry(ChitchatIdNodeEq(chitchat_id.clone()))
+            .entry(chitchat_id.clone())
             .or_insert_with(|| {
                 SamplingWindow::new(
                     self.config.sampling_window_size,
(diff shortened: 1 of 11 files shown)


trinity-1686a:
while trying to fix the tests, I found 2 issues in a test which worked before, as far as I can tell because of favorable timing.
- after a partition, when recovery start, if a node used to have a key, but has deleted it and has none left, other nodes may not learn about its max_version, and accept stall writes from nodes that aren't yet reset. It itself won't reset further, as it now has up to date heartbeat => send a sentinel value (tombstone for "") if we reset a node, and gives it no key at all.
- if a node A thinks an other node B is dead, it no longer includes it in its digest, which means other nodes will tell it about a new version of B, but won't tell it to reset its state. If B created a key before disconnection, deleted it during that phase, and its tombstone expired during that phase too, when A and B starts to communicate again, A will get an up to date heartbeat, but will keep the old key for A as neither tombstone nor state reset told it to drop the key => always include dead nodes in digest so we get resets when we should

chitchat/src/failure_detector.rs
```diff
@@ -181,10 +194,14 @@ impl SamplingWindow {
 
     /// Computes the sampling window's phi value.
     pub fn phi(&self) -> f64 {
-        // Ensure we don't call before any sample arrival.
-        assert!(self.intervals.mean() > 0.0 && self.last_heartbeat.is_some());
-        let elapsed_time = self.last_heartbeat.unwrap().elapsed().as_secs_f64();
-        elapsed_time / self.intervals.mean()
+        if self.last_heartbeat.is_none() {
```
guilload, line 197: nit: Can we use an `if let Some(...)` or `let Some(..)` here? That way, I don't have to think about whether the call to `unwrap` is safe.

chitchat/src/state.rs
```diff
@@ -366,21 +365,39 @@ impl ClusterState {
...
             }
+            let mut added_something = false;
             for (key, versioned_value) in stale_node.stale_key_values() {
+                added_something = true;
                 if !delta_writer.add_kv(key, versioned_value.clone()) {
                     let delta: Delta = delta_writer.into();
                     return delta;
                 }
             }
+            if !added_something && nodes_to_reset.contains(&stale_node.chitchat_id) {
```
guilload, line 385: I'd rather name the key `reset_sentinel` or something like that for debugging.

chitchat/src/state.rs
```diff
@@ -366,21 +365,39 @@ impl ClusterState {
...
+                // send a sentinel element to update the max_version. Otherwise the node's vision
+                // of max_version will be 0, and it may accept writes that are supposed to be
+                // stale, but it can tell they are.
+                if !delta_writer.add_kv(
```
guilload, line 389: This branch does not seem unit-tested.

fulmicoton:
```
what I implemented to fix that is that on first learning about a node, it's recorded in the failure detector as dead. if we get a 2nd heartbeat (with higher sequence number), we update to alive.
```

I think it is possible for a Node 4 to send Node 3 info about Node 2 with a higher sequence number. Node 3 would then update it as alive for a little bit. Can you confirm that this scenario will just result in Node 3 thinking for a few seconds that Node 2 is alive.

fulmicoton (approved):

fulmicoton:
Can we merge this? If you have identified more faulty corner case, please open an issue to document them.

PR #112, opened 2024-02-12
Compression

fulmicoton:
Adding ZSTD compression to chitchat's Deltas.

    The difficulty in this PR is coming from the fact that
    we need to observe the maximum transmissible unit,
    and due to the fact that we want to keep the simplficity
    of using the same `MessageDigest` object when
    sending and emitting messages.

    The solution I went for requires to compress things twice.
    When building our message object, we rely on a stream
    compressor that makes it possible to get an upperbound of the
    resulting payload after we insert an arbitrary serializable object.

    We then decompose the delta into as many mutations.

    I had to refactor the delta object, so that we have a proper bijection
    between a delta object and the list of mutations that were used to
    build it.

    The delta object hence encodes the order in which nodes, and values
    were inserted.
    The point there is to ensure that the actual serialization will
    yield the same size as the "simulated one" and hence respect the mtu.

    The code is a bit simpler that way too.

### DIFF
--- a/chitchat/src/lib.rs
+++ b/chitchat/src/lib.rs
@@ -29,7 +29,6 @@ use tracing::{error, warn};
 pub use self::configuration::ChitchatConfig;
 pub use self::state::{ClusterStateSnapshot, NodeState};
 use crate::digest::Digest;
-use crate::message::syn_ack_serialized_len;
 pub use crate::message::ChitchatMessage;
 pub use crate::server::{spawn_chitchat, ChitchatHandle};
 use crate::state::ClusterState;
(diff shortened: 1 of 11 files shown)


chitchat/src/lib.rs
```diff
@@ -159,16 +156,17 @@ impl Chitchat {
     /// Reports heartbeats to the failure detector for nodes in the delta for which we received an
     /// update.
     fn report_heartbeats(&mut self, delta: &Delta) {
-        for (chitchat_id, node_delta) in &delta.node_deltas {
-            if let Some(node_state) = self.cluster_state.node_states.get(chitchat_id) {
-                if node_state.heartbeat() < node_delta.heartbeat
-                    || node_state.max_version() < node_delta.max_version
```
fulmicoton, line 165: After this patch, we only record as heartbeat an update in the actual heartbeat KV.

chitchat/src/lib.rs
```diff
@@ -105,30 +109,24 @@ impl Chitchat {
...
-                    - syn_ack_serialized_len(&self_digest, &empty_delta);
-                let delta = self.cluster_state.compute_delta(
+                let delta_mtu = MAX_UDP_DATAGRAM_PAYLOAD_SIZE - 1 - digest.serialized_len();
+                let delta = self.cluster_state.compute_partial_delta_respecting_mtu(
                     &digest,
                     delta_mtu,
                     &dead_nodes,
                     self.config.marked_for_deletion_grace_period as u64,
                 );
-                self.report_heartbeats(&delta);
```
fulmicoton, line 120: I believe this was an error.

guilload (approved):

PR #116, opened 2024-02-15
Change in the liveness detection.

fulmicoton:
- Updating heartbeat on Digest messages (in addition to process delta)
- Avoid making up an interval when we have received a single heartbeat.
- Never marking nodes with less than 2 heartbeat recorded as live.
- Using additive smoothing to deal with the problem of the inability to assess liveness on the first few samples.

### DIFF
--- a/chitchat/src/failure_detector.rs
+++ b/chitchat/src/failure_detector.rs
@@ -1,4 +1,5 @@
 use std::collections::{HashMap, HashSet};
+use std::num::NonZeroUsize;
 use std::time::Duration;
 #[cfg(not(test))]
 use std::time::Instant;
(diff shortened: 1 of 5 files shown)


chitchat/src/failure_detector.rs
```diff
@@ -182,102 +185,107 @@ struct SamplingWindow {
...
     /// Returns the mean.
-    pub fn mean(&self) -> f64 {
-        self.mean
+    pub fn sum(&self) -> f64 {
+        self.sum
     }
 
-    /// Appends a new value and updates the statistics.
-    pub fn append(&mut self, interval: f64) {
-        if self.index == self.size {
```
fulmicoton, line 262: I really dislike this kind of code because it is impossible to put a meaning to `self.index`.

PR #120, opened 2024-02-19
Change in chitchat gossiping priority.

fulmicoton:
Following the original paper, chitchat currently first shares nodes with the highest number of stale values.

As a side effect, nodes that are not emitting many KVs are gossiped last. In quickwit, under a little bit of load (1000 indexes on 10 indexer), it has a very dire effect.

Indexer that reconnect have to gossip the entire cluster state (~10MB) before being able to get any information about the metastore.
Knowing at least one node with the metastore service is required for nodes to declare themselves as live.

This PR changes the gossip order.
It prioritizes nodes for which the node that originated the Syn message has not receives any KV yet.

This is identified by the fact that either the node was not part of the digest at all, or the floor_version is equal to 0.

The latter (floor_version == 0) should not happen today, but this case is done in preparation for another PR updating heartbeat on Syn.

### DIFF
--- a/chitchat/src/state.rs
+++ b/chitchat/src/state.rs
@@ -1,3 +1,4 @@
+use std::cmp::Ordering;
 use std::collections::btree_map::Entry;
 use std::collections::{BTreeMap, HashSet};
 use std::fmt::{Debug, Formatter};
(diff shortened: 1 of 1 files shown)


chitchat/src/state.rs
```diff
@@ -402,20 +403,77 @@ impl ClusterState {
...
 }
 
+/// The `staleness_score` is used to decide which node should be gossiped first.
+/// `floor_version` is the version (transmitted in the digest), below which
+/// all of the records have already been received.
+///
+/// There is no such thing as a KV for version 0. So if `floor_version == 0`
+/// it means the node is entirely new.
+/// We artificially prioritize those nodes to make sure their membership (in quickwit the service
+/// key for instnace) and initial KVs spread rapidly.
```
fmassot, line 448:
```suggestion
/// key for instance) and initial KVs spread rapidly.
```

chitchat/src/state.rs
```diff
@@ -402,20 +403,77 @@ impl ClusterState {
...
+/// If no KV is stale, there is nothing to gossip, and we simply return `None`:
+/// the node is not a candidate for gossip.
+fn staleness_score(node_state: &NodeState, floor_version: u64) -> Option<Staleness> {
+    let is_unknown = floor_version == 0u64;
+    let num_stale_kv = if is_unknown {
+        node_state.num_key_values()
+    } else {
+        node_state.stale_key_values(floor_version).count()
+    };
+    // We don't return None if the node is unknown yet because we want to make
```
fmassot, line 459: the comment is not very clear. The function comment is clearer so I would remove this one

chitchat/src/state.rs
```diff
@@ -402,20 +403,77 @@ impl ClusterState {
...
 
+/// Score used to decide which member should be gossiped first.
+///
 /// Number of stale key-value pairs carried by the node. A key-value is considered stale if its
 /// local version is higher than the max version of the digest, also called "floor version".
-type Staleness = usize;
+#[derive(Clone, Copy, Eq, PartialEq)]
+struct Staleness {
+    is_unknown: bool,
+    num_stale_records: usize,
```
guilload, line 413:
```suggestion
    num_stale_key_values: usize,
```

chitchat/src/state.rs
```diff
@@ -402,20 +403,77 @@ impl ClusterState {
...
+}
+
+/// The ord should be considered a "priority". The higher the faster a node's
```
guilload, line 416:
```suggestion
/// The ord should be considered a "priority". The higher, the faster a node's
```

chitchat/src/state.rs
```diff
@@ -402,20 +403,77 @@ impl ClusterState {
...
+/// information is gossiped.
+impl Ord for Staleness {
+    fn cmp(&self, other: &Self) -> Ordering {
+        // Nodes get gossiped in priority.
```
guilload, line 420:
```suggestion
        // Nodes are gossiped in order of priority:
```

chitchat/src/state.rs
```diff
@@ -402,20 +403,77 @@ impl ClusterState {
...
 /// Sorts the stale nodes in decreasing order of staleness.
-#[derive(Debug, Default)]
+#[derive(Default)]
 struct SortedStaleNodes<'a> {
     stale_nodes: BTreeMap<Staleness, Vec<StaleNode<'a>>>,
 }
 
+/// The `staleness_score` is used to decide which node should be gossiped first.
+/// `floor_version` is the version (transmitted in the digest), below which
+/// all of the records have already been received.
```
guilload, line 443:
```suggestion
/// all the records have already been received.
```

chitchat/src/state.rs
```diff
@@ -402,20 +403,77 @@ impl ClusterState {
...
+///
+/// There is no such thing as a KV for version 0. So if `floor_version == 0`
```
guilload, line 445:
```suggestion
/// There is no such thing as a KV for version 0. So if `floor_version == 0`,
```

PR #122, opened 2024-02-19
Key deletion logic bugfix

### LINKED ISSUE #121: Key GC logic is broken
The current key GC logic is broken. It relies on heartbeat, assuming that all keys that were updated before the last heartbeat have been received. This is however incorrect.

The heartbeat is sent with every single delta.
If a delta of a node is partial  (meaning it does not contain all of the stale KV of a node), the delta receiver might update its heartbeat without receiving some anterior KV updates.

For instance, the following scenario leads into KV reappearing after deletion.
```
- node1 inserts key K
- key K is replicated to node 2
- node1 deletes key K at version V
- some time passes
- node2 receives partial update not contain key K but updating Node 1 heartbeat.
- grace period is elapsed
- node1 GC key K
```

More than just being a bug, relying on heartbeat like this prevents us from using heartbeat sent in digest to discover and
detect which nodes is alive.

fulmicoton:
Closes #121


chitchat/src/state.rs
```diff
@@ -320,15 +338,8 @@ impl ClusterState {
         }
     }
 
-    pub fn gc_keys_marked_for_deletion(
-        &mut self,
-        marked_for_deletion_grace_period: u64,
-        dead_nodes: &HashSet<ChitchatId>,
-    ) {
-        for (chitchat_id, node_state) in &mut self.node_states {
```
fulmicoton, line 328: No clue why we skipped dead nodes in GC.

README.md
```diff
@@ -56,23 +56,25 @@ associated with a version, and replicated using the same mechanism as a KV updat
 The library will then interpret this versioned tombstone before exposing kv to
 the user.
 
-To avoid keeping deleted KV indefinitely, the library includes a GC mechanism. Any nodes containing a tombstone older than a given `grace period threshold`
-(age is measured in ticks of heartbeat), it is safe to be deleted.
+To avoid keeping deleted KV indefinitely, the library includes a GC mechanism.
+Every tombstone is associated with a monotonic timestamp.
+It is local in the sense that it is computed locally to the given node, and never shared with other servers.
+
+All KV with a timestamp older than a given `grace_period_threshold` will be deleted upon delete operations. (Note for a given KV, GC can happen at different
```
fmassot, line 63:
```suggestion
All KV with a timestamp older than a given `marked_for_deletion_grace_period` will be deleted upon delete operations. (Note for a given KV, GC can happen at different
```

chitchat/src/configuration.rs
```diff
@@ -56,7 +56,8 @@ impl Default for ChitchatConfig {
             failure_detector_config: Default::default(),
             // Each heartbeat increments the version, with one heartbeat each second
```
fmassot, line 57: you can remove the old comment

chitchat/src/state.rs
```diff
@@ -24,6 +26,11 @@ pub struct NodeState {
     max_version: Version,
     #[serde(skip)]
     listeners: Listeners,
+    // This is the maximum version of the last tombstone GC.
+    //
+    // After we GC the tombstones, we cannot safely do replication with
+    // nodes that are asking for a diff from with version lower than this.
```
fmassot, line 32:
```suggestion
    // nodes that are asking for a diff with version lower than this.
```

chitchat/src/configuration.rs
```diff
@@ -56,7 +56,8 @@ impl Default for ChitchatConfig {
...
             // 86400 ~ 24h.
-            marked_for_deletion_grace_period: 86400,
+            // TODO set that to something much lower.
```
fmassot, line 59: let's do it, something like 1 or 2 hours?

fmassot (approved):

PR #123, opened 2024-02-19
Use the heartbeats shipped with the digest for cluster membership and failure detection

fulmicoton:


chitchat/src/failure_detector.rs
```diff
@@ -55,23 +41,30 @@ impl FailureDetector {
...
+            .phi(chitchat_id)
+            .map(|phi| phi <= self.config.phi_threshold)
+            .unwrap_or(false);
+        debug!(node_id=%chitchat_id.node_id, phi=?phi_opt, is_alive=is_alive, "computing node liveness");
+        if is_alive {
+            self.live_nodes.insert(chitchat_id.clone());
+            self.dead_nodes.remove(chitchat_id);
+        } else {
+            self.live_nodes.remove(chitchat_id);
+            if !self.dead_nodes.contains_key(&chitchat_id) {
```
fulmicoton, line 61: this was a bug.
When a node was identified twice as dead, we would update its time of death.

chitchat/src/state.rs
```diff
@@ -303,22 +317,40 @@ impl ClusterState {
 
     pub(crate) fn apply_delta(&mut self, delta: Delta) {
         // Remove nodes to reset.
+        if delta.nodes_to_reset.len() > 0 {
+            tracing::error!(nodes_to_reset=?delta.nodes_to_reset, "nodes to reset");
```
fmassot, line 321: Is having to reset nodes an error? info or warning would be better

chitchat/src/state.rs
```diff
@@ -303,22 +317,40 @@ impl ClusterState {
...
                 key_values,
             } = node_delta;
-            let node_state = self
-                .node_states
-                .entry(chitchat_id.clone())
-                .or_insert_with(|| NodeState::new(chitchat_id, self.listeners.clone()));
-            if node_state.heartbeat < heartbeat {
-                node_state.heartbeat = heartbeat;
+            let node_state = self.node_state_mut(&chitchat_id);
+            if node_state.last_gc_version == 0u64 {
```
fmassot, line 334: Very helpful explanation.

chitchat/src/state.rs
fulmicoton: yes that was a mistake. I made it an info.

PR #124, opened 2024-02-20
Sample cluster state from peer seeds

guilload:
The idea is to allow fetching a few attributes (readiness, gRPC advertise address) from the peer seeds using `sample_from_seeds` to then download the cluster state directly via gRPC calls, and finally bootstrap our own state using `insert_or_create_node`.

### DIFF
--- a/chitchat/src/lib.rs
+++ b/chitchat/src/lib.rs
@@ -13,21 +13,29 @@ mod state;
 pub mod transport;
 mod types;
 
+use std::collections::hash_map::Entry;
 use std::collections::{BTreeMap, HashMap, HashSet};
 use std::iter::once;
 use std::net::SocketAddr;
+use std::time::Duration;
 
+use anyhow::bail;
 use delta::Delta;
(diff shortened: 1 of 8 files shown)


chitchat/src/lib.rs
```diff
@@ -85,6 +91,73 @@ impl Chitchat {
...
+        max_samples: u16,
+        include_keys: Vec<String>,
+    ) -> anyhow::Result<HashMap<ChitchatId, HashMap<String, VersionedValue>>> {
+        // Using a random UDP port for one-time use.
+        let mut socket = transport
+            .open(SocketAddr::new(gossip_listen_addr.ip(), 0))
+            .await?;
+
+        let mut seed_addrs: Vec<SocketAddr> = self.cluster_state.seed_addrs().into_iter().collect();
+        seed_addrs.shuffle(&mut rand::thread_rng());
```
guilload, line 109: If the seeds are empty, we could try from `self`. WDYT?

PR #128, opened 2024-02-23
Trigger catch-up callback when node is lagging behind

guilload:
### Todo
- [x] add unit test

### DIFF
--- a/chitchat/src/lib.rs
+++ b/chitchat/src/lib.rs
@@ -24,7 +24,7 @@ pub use listener::ListenerHandle;
 pub use serialize::Serializable;
 use tokio::sync::watch;
 use tokio_stream::wrappers::WatchStream;
-use tracing::{error, warn};
+use tracing::{error, info, warn};
 
 pub use self::configuration::ChitchatConfig;
 pub use self::state::{ClusterStateSnapshot, NodeState};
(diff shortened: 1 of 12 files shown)


chitchat/src/lib.rs
```diff
@@ -277,6 +294,20 @@ impl Chitchat {
         ClusterStateSnapshot::from(&self.cluster_state)
     }
 
+    pub fn insert_or_update_node(
+        &mut self,
+        chitchat_id: &ChitchatId,
+        key_values: impl Iterator<Item = (String, VersionedValue)>,
+    ) -> Version {
+        // TODO: How to insert in failure detector without reporting a heartbeat?
```
fulmicoton, line 302: I don't think this reports a heartbeat here.

fulmicoton: You need to update `last_gc_version` too.

chitchat/src/lib.rs
```diff
@@ -277,6 +294,20 @@ impl Chitchat {
...
+        let node_state = self.cluster_state.node_state_mut(chitchat_id);
+
+        for (key, value) in key_values {
+            node_state.set_versioned_value(key, value)
```
fulmicoton, line 306: I don't think it is correct to just insert. Some keys might need to be deleted.

fulmicoton (approved):

PR #130, opened 2024-02-26
Issue/128 delta validation

fulmicoton:


chitchat/src/server.rs
```diff
@@ -578,6 +578,12 @@ mod tests {
         let server_handle = spawn_chitchat(server_config, Vec::new(), &transport)
             .await
             .unwrap();
+        server_handle
+            .chitchat()
+            .lock()
+            .await
+            .self_node_state()
+            .set("key", "value");
```
RaphaelMarinier, line 586: Is this a leftover for a local test?

chitchat/src/state.rs
```diff
@@ -104,6 +119,122 @@ impl NodeState {
...
+        }
+
+        if self.max_version > node_delta.last_gc_version {
+            // The GCed tombstone have all been already received.
+            // We won't miss anything by applying the delta!
+            return true;
+        }
+
+        // This delta might be missing tombstones with version within
+        // (`node_state.max_version`..`last_gc_version`].
```
RaphaelMarinier, line 153: last_gc_version -> for clarity would be good to say whether this is self.last_gc_version or node_delta.last_gc_version.

chitchat/src/state.rs
```diff
@@ -24,14 +24,29 @@ pub struct NodeState {
...
+    // because a node that was just reset or just joined the cluster will get updates
+    // from another node that are actually only make sense in the context of the
+    // emission of delta from a `last_gc_version`.
     last_gc_version: Version,
+    // A proper interpretation of `max_version` and `last_gc_version` is the following:
+    // The state contains exactly:
+    // - all of the (non-deleted) key values present at snapshot `max_version`.
+    // - all of the tombstones of the entry that were deleted between (`last_gc_version`,
+    //   `max_version]`.
+    //
```
RaphaelMarinier, line 47: Can we add that we always have last_gc_version <= max_version?

chitchat/src/state.rs
```diff
@@ -104,6 +119,122 @@ impl NodeState {
...
+            // We won't miss anything by applying the delta!
+            return true;
+        }
+
+        // This delta might be missing tombstones with version within
+        // (`node_state.max_version`..`last_gc_version`].
+        //
+        // It is ok if we don't have the associated values to begin
+        // with.
+        if self.last_gc_version >= node_delta.last_gc_version {
```
RaphaelMarinier, line 157: This will almost never be true given that self.max_version >= self.last_gc_version (so self.last_gc_version >= node_delta.last_gc_version implies self.max_version >= node_delta_last_gc_version so the condition of the previous if will most of the times be true), right?
It seems that it would only be true if self.max_version == self.last_gc_version == node_delta.last_gc_version.

chitchat/src/state.rs
```diff
@@ -104,6 +119,122 @@ impl NodeState {
...
+            .max()
+            .or(node_delta.max_version)
+        else {
+            // This can happen if we just hit the mtu at the moment
+            // of writing the SetMaxVersion operation.
+            return false;
+        };
+
+        if delta_max_version <= self.max_version() {
+            // There is not point apply in this delta as it is not bringing us to a newer state.
```
RaphaelMarinier, line 183:
```suggestion
            // There is not point applying this delta as it is not bringing us to a newer state.
```

chitchat/src/state.rs
```diff
@@ -104,6 +119,122 @@ impl NodeState {
...
+
+        // We are out of sync. This delta is an invitation to `reset` our state.
+        info!(
+            node=?node_delta.chitchat_id,
+            last_gc_version=node_delta.last_gc_version,
+            current_last_gc_version=self.last_gc_version,
+            "resetting node");
+        *self = NodeState::new(node_delta.chitchat_id.clone(), self.listeners.clone());
+        if let Some(max_version) = node_delta.max_version {
+            self.max_version = max_version;
```
RaphaelMarinier, line 202: This relies on the fact that node_delta.max_version is only set when when node_delta does not contain any kv mutations (otherwise, apply_delta() would skip the kv mutations since they would be <= to the max_version we just set).
Maybe comment this somewhere, e.g. in NodeDelta?
It may be a bit brittle, since setting NodeDelta.max_version even though it's not necessary would break this, but hey.

chitchat/src/state.rs
```diff
@@ -24,14 +24,29 @@ pub struct NodeState {
...
+    // `last_gc_version` expresses the idea: what is the oldest version from which I can
+    // confidently emit delta from. The reason why we update it here, is
+    // because a node that was just reset or just joined the cluster will get updates
+    // from another node that are actually only make sense in the context of the
+    // emission of delta from a `last_gc_version`.
     last_gc_version: Version,
+    // A proper interpretation of `max_version` and `last_gc_version` is the following:
+    // The state contains exactly:
+    // - all of the (non-deleted) key values present at snapshot `max_version`.
+    // - all of the tombstones of the entry that were deleted between (`last_gc_version`,
```
RaphaelMarinier, line 45: Maybe clearer:
```suggestion
    // - all of the tombstones of the entry that were marked for deletion between (`last_gc_version`,
```

chitchat/src/state.rs
```diff
@@ -24,14 +24,29 @@ pub struct NodeState {
...
+    //   `max_version]`.
+    //
+    // It does not contain any trace of the tombstones of the entries that were deleted before
```
RaphaelMarinier, line 48:
```suggestion
    // It does not contain any trace of the tombstones of the entries that were marked for deletion before
```

chitchat/src/state.rs
```diff
@@ -421,70 +510,64 @@ impl ClusterState {
...
+                .get(chitchat_id)
+                .map(|node_digest| node_digest.max_version)
+                .unwrap_or(0u64);
+
+            if node_state.max_version <= digest_max_version {
+                // Our version is actually older than the version of the digest.
+                // We have no update to offer.
                 continue;
-            };
-            // TODO: We have problem here. If after the delta we end up with a max version that is
```
RaphaelMarinier, line 434: FYI, I believe that this problem is still present in the new version, which can lead to a node always being reset and never catching up to the max_version of the other nodes.

e.g:
Node A has last_gc=1000, many kvs in [0, 2000], and max_version=2000.
Node B was disconnected and has max_version=500.
B sends a SYN with max_version=500, A decides that it should send "should_reset" delta with last_gc=1000, from_version=0, and all the kvs in [0, 2000].
Now, if all the kvs that fit in the message have versions < 1000, B will only apply thoses deltas up to a version < 1000, and have max_version<1000. The next time it sends its digest to A, A will again decide it should send a "should_reset" delta and again send the kvs starting at 0.

The same can happen if A only has non-removed kvs in the range [0, 999].

Also, since "should_reset" deltas are sent first in the packet based on the scuttle-depth heuristic, if they reach the size of the packet, those should_reset updates sent to B will also prevent indefinitely updates to the kvs in B's cluster state for the other nodes.

So maybe preserve/extend the original comment somewhere?

chitchat/src/delta.rs
```diff
@@ -83,7 +84,6 @@ impl TryFrom<u8> for DeltaOpTag {
         match tag_byte {
             0u8 => Ok(DeltaOpTag::Node),
             1u8 => Ok(DeltaOpTag::KeyValue),
-            2u8 => Ok(DeltaOpTag::NodeToReset),
             _ => {
```
RaphaelMarinier, line 87: You will need to handle 2u8 for SetMaxVersion here.

chitchat/src/state.rs
fulmicoton: No... But I can add the opposite because this is not true :)

chitchat/src/state.rs
fulmicoton: It is possible to have max_version < last_gc_version.

If a node gets reset, it will set its last_gc_version to something else, and will receive a partial delta.

chitchat/src/state.rs
fulmicoton: good point!

chitchat/src/delta.rs
fulmicoton: Great point!

chitchat/src/state.rs
```diff
@@ -421,70 +510,64 @@ impl ClusterState {
...
-                        version: stale_node.node_state.max_version,
-                        tombstone: Some(Instant::now()),
-                    },
-                ) {
-                    return delta_serializer.finish();
-                }
+            // There aren't any key-values in the state_node apparently.
+            // Let's add a specific instruction to the delta to set the max version.
+            if !added_something {
+                delta_serializer.try_set_max_version(stale_node.node_state.max_version);
```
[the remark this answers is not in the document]
fulmicoton: Good point. I also added a #[must_use] to the method.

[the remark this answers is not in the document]
fulmicoton: yes. I just added a comment.

chitchat/src/state.rs
fulmicoton: You are absolutely right. I also noticed the problem when writing down what I thought the invariant were and made an update.

In a nutshell, we do no need to reset if `digest_last_gc_version >= self.last_gc_version`.

chitchat/src/state.rs
```diff
@@ -421,70 +510,64 @@ impl ClusterState {
...
-                    "__reset_sentinel",
-                    VersionedValue {
-                        value: String::new(),
-                        version: stale_node.node_state.max_version,
-                        tombstone: Some(Instant::now()),
-                    },
-                ) {
-                    return delta_serializer.finish();
-                }
+            // There aren't any key-values in the state_node apparently.
```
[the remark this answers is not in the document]
fulmicoton: yes.

ALGORITHM.md
```diff
@@ -0,0 +1,66 @@
...
+
+- `max_version`: One big difference with reliable broadcast, is that this `max version` does NOT mean that the node's state reflects all of the events earlier than
+`max_version`. Rather, it says that if the node were to receive all of the events within between `max_version` and $V$ then it will end up to date.
+The difference is subtle, but comes from an important optimization of scuttlebutt.
+If a node ask for an update above $max_version = m$, and that for a given key $k$ there are two updates after $m$, there is no need to send the first one: it will eventually be overridden by the second one.
+
+- `last_gc_version`: Rather than being deleted right away, the algorithm is using a tombstone mechanism. Key entries are kept by marked as deleted. From the client point of view, everything looks like the key really has been deleted. Eventually, in order
+to free memory, we garbage collect all of the key values marked for deletion that are older than a few hours. We then keep track of the maximum `last_gc_version`.
+
+At each round of the protocol, a node makes sure to its pair $(last\_gc\_version, max\_version)$ in the lexicographical sense.
```
RaphaelMarinier, line 42:
```suggestion
At each round of the protocol, a node maintains a pair $(last\_gc\_version, max\_version)$ which can only increase in lexicographical order.
```

RaphaelMarinier (approved):

chitchat/src/state.rs
fulmicoton: Yeah this is way too brittle. I'll change this to make it into a max.

PR #137, opened 2024-03-07
Added a way to set a TTL to key values.

fulmicoton:
After the GC grace period, the key value will be subject to the GC just like regularly deleted KVs.

### DIFF
--- a/chitchat/src/state.rs
+++ b/chitchat/src/state.rs
@@ -17,6 +17,7 @@ use tracing::{info, warn};
 use crate::delta::{Delta, DeltaSerializer, NodeDelta};
 use crate::digest::{Digest, NodeDigest};
 use crate::listener::Listeners;
+use crate::types::{DeletionStatus, DeletionStatusMutation};
 use crate::{ChitchatId, Heartbeat, KeyChangeEvent, Version, VersionedValue};
 
 #[derive(Clone, Serialize, Deserialize)]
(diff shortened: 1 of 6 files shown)


chitchat/src/state.rs
```diff
@@ -1051,12 +1076,17 @@ mod tests {
...
         node_state.set("key", "1");
-        node_state.mark_for_deletion("key");
+        node_state.delete("key");
+        assert!(node_state.get("key").is_none());
         {
             let versioned_value = node_state.get_versioned("key").unwrap();
             assert_eq!(&versioned_value.value, "");
             assert_eq!(versioned_value.version, 2u64);
-            assert!(&versioned_value.is_tombstone());
+            assert!(&versioned_value.is_deleted());
```
trinity-1686a, line 1085:
```suggestion
            assert!(versioned_value.is_deleted());
```
i don't think the borrow is necessary, and i find it somewhat confusing (is it supposed to be a borrow, or a mistyped !)

chitchat/src/types.rs
```diff
@@ -108,7 +133,7 @@ impl PartialEq for VersionedValue {
     fn eq(&self, other: &Self) -> bool {
         self.value.eq(&other.value)
             && self.version.eq(&other.version)
-            && self.is_tombstone().eq(&other.is_tombstone())
+            && self.is_deleted().eq(&other.is_deleted())
```
trinity-1686a, line 136: I'm not sure is_deleted is enough to qualify equality. `DeletionStatus::Set` and `DeletionStatus::DeleteAfterTtl` are considered equal by this statement. If that is intended, a comment should explain why

trinity-1686a (commented):
looks sensible; see comments

chitchat/src/state.rs
fulmicoton: yes!

chitchat/src/types.rs
fulmicoton: Thx. I switched to checking the discriminant, added a comment, and made PartialEq cfg test only.

PR #146, opened 2024-03-30
Replace `return` with `warn!` in server loop

### LINKED ISSUE #145: Weird behavior when running tests
I'm running the tests from "chitchat-test" locally on my windows machine with the following script (trying to mimic the bash version as best I can): 

```
Get-Process "chitchat-test" | Stop-Process

cargo build --release

for ($i = 10000; $i -lt 10100; $i++)
{
    $listen_addr = "127.0.0.1:$i";
    Write-Host $listen_addr;

    Start-Process -NoNewWindow "cargo" -ArgumentList "run --release -- --listen_addr $listen_addr --seed 127.0.0.1:10002 --node_id node_$i"
}

Read-Host
```

Are the following results expected behavior, or am I maybe running into some windows bugs?

1. The services started at 127.0.0.1:10000 and 127.0.0.1:10001 (before the seed address port) never receive any peers, and are stuck with a rather bare "state", regardless of how long I leave it running:

```
{
  "cluster_id": "testing",
  "cluster_state": {
    "node_state_snapshots": [
      {
        "chitchat_id": {
          "node_id": "node_10000",
          "generation_id": 1711736008,
          "gossip_advertise_addr": "127.0.0.1:10000"
        },
        "node_state": {
          "chitchat_id": {
            "node_id": "node_10000",
            "generation_id": 1711736008,
            "gossip_advertise_addr": "127.0.0.1:10000"
          },
          "heartbeat": 2,
          "key_values": {},
          "max_version": 0,
          "last_gc_version": 0
        }
      }
    ],
    "seed_addrs": [
      "127.0.0.1:10002"
    ]
  },
  "live_nodes": [
    {
      "node_id": "node_10000",
      "generation_id": 1711736008,
      "gossip_advertise_addr": "127.0.0.1:10000"
    }
  ],
  "dead_nodes": []
}
```

The other services on ports greater than the seed start up fine.

2. If I select a random process and kill it, all heartbeats stop (even for the services which received no peers).  The state is still readable as JSON with a GET request to 127.0.0.1:10XXX, but the heartbeat does not increment, and no new peers are registered if I try adding them after the stoppage.

Is any of this expected behavior, or do we think maybe its windows related?

cattuz:
The return causes the server to stop and all gossiping to end whenever a node leaves on windows. Each node which fails to communicate with the exited node will then stop gossiping.

Attempts to fix #145

### DIFF
--- a/chitchat-test/run-servers.ps1
+++ b/chitchat-test/run-servers.ps1
@@ -1,0 +1,13 @@
+Get-Process "chitchat-test" | Stop-Process
+
+cargo build --release
+
+for ($i = 10000; $i -lt 10100; $i++)
+{
+    $listen_addr = "127.0.0.1:$i";
+    Write-Host $listen_addr;
+
+    Start-Process -NoNewWindow "cargo" -ArgumentList "run --release -- --listen_addr $listen_addr --seed 127.0.0.1:10002 --node_id node_$i"
+}
+
+Read-Host
--- a/chitchat/src/server.rs
+++ b/chitchat/src/server.rs
@@ -7,7 +7,7 @@ use rand::prelude::*;
 use tokio::net::lookup_host;
 use tokio::sync::mpsc::{self, UnboundedReceiver, UnboundedSender};
 use tokio::sync::{watch, Mutex};
-use tokio::task::JoinHandle;
+use tokio::task::{JoinError, JoinHandle};
 use tokio::time;
 use tracing::{debug, info, warn};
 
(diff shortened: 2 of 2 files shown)


fulmicoton-dd:
@rdettai can you have a look? I agree it is weird to exit the loop when receiving a message.

fulmicoton-dd:
@copilot resolve the merge conflicts in this pull request

fulmicoton:
@cattuz thanks for the report.

I merged your fixed in main in a commit that is attributed to you... And reverting right now to address it elsewhere.

If you have a look at the Socket spec.

```
#[async_trait]
pub trait Socket: Send + Sync + 'static {
    // Only returns an error if the transport is broken and may not emit message
    // in the future.
    async fn send(&mut self, to: SocketAddr, msg: ChitchatMessage) -> anyhow::Result<()>;
    // Only returns an error if the transport is broken and may not receive message
    // in the future.
    async fn recv(&mut self) -> anyhow::Result<(SocketAddr, ChitchatMessage)>;
}
```

It is in charge of classifying transient errors, and not surface them to the caller.
I also fixed the classifier so we detect your windows errors.

PR #148, opened 2024-04-30
Batch subscriber

fulmicoton:
Upon reset_node and apply_update, we want to be
able to react to a batch of changes rather than individual changes.

In quickwit for instance, we want to be able to react to the reception of a batch of deleted shard and group this reaction into a single metastore call.

### DIFF
--- a/chitchat/src/lib.rs
+++ b/chitchat/src/lib.rs
@@ -22,6 +22,7 @@ use failure_detector::FailureDetector;
 pub use failure_detector::FailureDetectorConfig;
 pub use listener::ListenerHandle;
 pub use serialize::Serializable;
+pub use state::NodeStateMut;
 use tokio::sync::watch;
 use tokio_stream::wrappers::WatchStream;
 use tracing::{error, info, warn};
@@ -31,7 +32,7 @@ pub use self::state::{ClusterStateSnapshot, NodeState};
 use crate::digest::Digest;
 pub use crate::message::ChitchatMessage;
 pub use crate::server::{spawn_chitchat, ChitchatHandle};
-use crate::state::ClusterState;
+pub use crate::state::ClusterState;
 pub use crate::types::{ChitchatId, DeletionStatus, Heartbeat, Version, VersionedValue};
 
 /// Maximum UDP datagram payload size (in bytes).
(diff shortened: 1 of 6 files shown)


chitchat/src/lib.rs
```diff
@@ -304,17 +299,39 @@ impl Chitchat {
...
     /// Existing key-values that are not present in `key_values` will be deleted
     /// (not marked with a tombstone).
+    ///
+    /// This method returns an error if the key values version are inconsistent with the max
+    /// version.
     pub fn reset_node_state(
         &mut self,
         chitchat_id: &ChitchatId,
-        key_values: impl Iterator<Item = (String, VersionedValue)>,
+        key_values: Vec<(String, VersionedValue)>,
```
fulmicoton, line 308: I added a defensive check in this method

chitchat/src/state.rs
```diff
@@ -822,6 +773,213 @@ impl From<&ClusterState> for ClusterStateSnapshot {
...
+        self.node_state_mut
+    }
+}
+
+impl<'a> DerefMut for NodeStateMut<'a> {
+    fn deref_mut(&mut self) -> &mut NodeState {
+        self.node_state_mut
+    }
+}
+
```
trinity-1686a, line 796: I think deref makes sense, but deref_mut is bound to cause mutations which don't notify listeners

chitchat/src/state.rs
```diff
@@ -822,6 +773,213 @@ impl From<&ClusterState> for ClusterStateSnapshot {
...
+            {
+                // No need to change anything, the value is already set!
+                return;
+            }
+        }
+        let new_version = self.node_state_mut.max_version + 1;
+        self.set_versioned_value(
+            key.to_string(),
+            VersionedValue {
+                value: value.to_string(),
```
trinity-1686a, line 904: these to_string() look redundant

trinity-1686a (commented):
i still have to read through half of state.rs. In the meantime, here are my first comments

chitchat/src/state.rs
```diff
@@ -961,49 +1140,72 @@ mod tests {
...
-        node_state4.set_with_version("key_b", "value_b", 2);
-        node_state4.set_with_version("key_c", "value_c", 5);
-        node_state4.set_with_version("key_d", "value_d", 7);
-        stale_nodes.offer(&node4, &node_state4, 1);
+        // 0 stale values
+        {
+            cluster_state.node_state_mut(&node5);
+        }
 
         // 3 stale values
```
trinity-1686a, line 1189: I don't understand where the stale value count come from. Before node5 and node6 had respectively 3 and 0 (instead of 0 and 3) stale values. I don't see an obvious reason this changed

trinity-1686a (changes requested):
(see also first batch)

PR #155, opened 2025-03-04
Fix dead node escaping deletion

rdettai:
In the Airmail Quickwit deployment we see situations where the cluster state is never cleaned up.

This PR fixes an edge case where multiple nodes flag a node as deleted at different times, which means that it is possible to receive a dead node's state again after its garbage collection.


chitchat/src/lib.rs
```diff
@@ -297,21 +323,23 @@ impl Chitchat {
...
+    /// A node that doesn't exist will be created.
     pub fn reset_node_state(
         &mut self,
         chitchat_id: &ChitchatId,
         key_values: impl Iterator<Item = (String, VersionedValue)>,
         max_version: Version,
         last_gc_version: Version,
     ) {
-        let node_state = self.cluster_state.node_state_mut(chitchat_id);
+        let node_state = self.cluster_state.node_state_mut_or_init(chitchat_id);
```
rdettai, line 342: I have some doubts about this function. If my understanding is correct, it is used in [Quickwit](https://github.com/quickwit-oss/quickwit/blob/de78b892bfa7f82546635e899fba4e999d3ba02c/quickwit/quickwit-cluster/src/grpc_gossip.rs) to provide a quick catchup of the state using GRPC. My concern would be that this catchup could be a backdoor that would prevent this fix from working by allowing dead nodes to come back into the state. I would otherwise go for something more conservative:
```rust
        let node_state = if self.cluster_state().last_heartbeat_if_deleted(chitchat_id).is_some() {
            self.cluster_state.node_state_mut_or_init(chitchat_id)
        } else if let Some(node_state) = self.cluster_state.node_state_mut(chitchat_id) {
            node_state
        } else {
            return;
        };
```

chitchat/tests/cluster_test.rs
```diff
@@ -122,6 +180,49 @@ impl Simulator {
...
+                                tokio::time::sleep(Duration::from_millis(100)).await;
+                            }
+                        }).await.map_err(|_| {
+                            anyhow!("Node {:?} assert timeout on chitchat_id={}", expected_status, chitchat_id.node_id)
+                        }).unwrap();
+                    } else {
+                        let chitchat_guard = chitchat.lock().await;
+                        info!(node_states=?chitchat_guard.node_states(), "Node presence assert");
+                        info!(live_node_state=?chitchat_guard.live_nodes().collect_vec(), "Node presence assert");
+                        info!(dead_node_state=?chitchat_guard.dead_nodes().collect_vec(), "Node presence assert");
```
fulmicoton, line 218: why 3 lines? you can have more than one structure param per info log line.

chitchat/src/types.rs
```diff
@@ -16,8 +16,7 @@ use crate::Serializable;
 /// The `generation_id` is used to detect when a node has restarted. It must be monotonically
 /// increasing to differentiate the most recent state and must be incremented every time a node
 /// leaves and rejoins the cluster. Backends such as Cassandra or Quickwit typically use the node's
-/// startup time as the `generation_id`. Applications with stable state across restarts can use a
-/// constant `generation_id`, for instance, `0`.
```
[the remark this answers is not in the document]
fulmicoton: What happens when facing a partition:
A node that is not dead ends up appearing dead for enough time to be GCed.

chitchat/src/types.rs
```diff
@@ -315,7 +317,10 @@ pub struct Heartbeat(pub(crate) u64);
 
 impl Heartbeat {
     pub(crate) fn inc(&mut self) {
-        self.0 = self.0.wrapping_add(1);
+        // Heartbeat is increased everytime we gossip or receive a message. In
+        // real-world scenarios, this should not happen more than a few hundred
+        // times per second so we should never overflow.
+        self.0 = self.0.checked_add(1).expect("Heartbeat overflow");
```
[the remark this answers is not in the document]
fulmicoton: It would take 100 years to wrap over with a 64 bits integer at 4GHz.

chitchat/tests/cluster_test.rs
rdettai: Aouch, I was meaning to remove these.

chitchat/src/types.rs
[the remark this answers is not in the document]
rdettai: It's handled by recording the last seen timestamp and tested here https://github.com/quickwit-oss/chitchat/blob/fix-surviving-node-bug/chitchat%2Ftests%2Fcluster_test.rs#L865-L881

.gitignore
```diff
@@ -3,7 +3,6 @@
 **/target/**
 **/proptest-regressions
 **/.idea
-Cargo.lock
```
fulmicoton-dd, line 6: why!?

Cargo.lock
```diff
@@ -0,0 +1,2766 @@
+# This file is automatically @generated by Cargo.
+# It is not intended for manual editing.
```
fulmicoton-dd, line 2: Why!?

fulmicoton-dd (approved):

chitchat/src/lib.rs
rdettai: to avoid that the catchup process recreates dead nodes again, recreating garbage collected node is disabled for `reset_node_state()`

PR #157, opened 2025-03-12
Avoid unstable const unwrap

rdettai:
Const panic is an unstable feature: https://rust-lang.github.io/rfcs/2345-const-panic.html

Quickwit CI fail: https://github.com/quickwit-oss/quickwit/actions/runs/13813433243/job/38640392942

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,10 @@ license-check: check-license-tool
 license-fix: check-license-tool
 	dd-rust-license-tool write
 
+build-compat: fmt
+	@echo "Running `cargo build --package chitchat` with the min compatible Rust version"
+	@RUSTUP_TOOLCHAIN=1.70 cargo build --package chitchat
+
 fix: fmt
 	@echo "Running cargo clippy --fix"
 	@cargo clippy --fix --all-features --allow-dirty --allow-staged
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -10,6 +10,10 @@ license-check: check-license-tool
 license-fix: check-license-tool
 	dd-rust-license-tool write
 
+build-compat: fmt
```
fulmicoton-dd, line 13: can we just add a rust-toolchain.toml file to the directory instead?

rdettai: Are you suggesting to set:
```toml
[toolchain]
channel = "1.70"
```
?

The problem is that the:
- tests use libraries that have more recent min rust-version (I even I saw one around 1.8x)
- the fmt rules that we expect are those of more recent versions

So we would need to set a fairly recent version of rust into `rust-toolchain.toml`, and that would defeat the purpose of validating the compatibility with older versions. This specific make rule is the only obvious way I found to try out a specific (old) version on just the library code.

fulmicoton-dd:
fixed in another PR. (I removed the cumbersome poem dependency by removing the demo main program)
rust version is set to 1.82, which is the same as quickwit.
I also updates the license file.

PR #162, opened 2025-06-25
Expose shutdown lifecycle on ChitchatHandle

rdettai:
To handle chitchat failures and panics in QW, we need to be able to watch its state (https://github.com/quickwit-oss/quickwit/pull/5819)

### DIFF
--- a/chitchat/src/server.rs
+++ b/chitchat/src/server.rs
@@ -1,4 +1,5 @@
 use std::collections::HashSet;
+use std::future::Future;
 use std::net::SocketAddr;
 use std::sync::Arc;
 use std::time::Duration;
@@ -28,6 +29,7 @@ pub struct ChitchatHandle {
     command_tx: UnboundedSender<Command>,
     chitchat: Arc<Mutex<Chitchat>>,
     join_handle: JoinHandle<Result<(), anyhow::Error>>,
+    termination_watcher: watch::Receiver<Option<Result<(), String>>>,
 }
 
 impl ChitchatHandle {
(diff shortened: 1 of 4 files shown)


chitchat/src/server.rs
```diff
@@ -186,6 +199,20 @@ impl ChitchatHandle {
         self.command_tx.send(Command::Gossip(addr))?;
         Ok(())
     }
+
+    pub fn termination_watcher(&self) -> impl Future<Output = anyhow::Result<()>> {
+        let mut watcher = self.termination_watcher.clone();
+        async move {
+            let termination_res = watcher.wait_for(|res| res.is_some()).await;
+            if let Ok(result_opt) = termination_res {
+                result_opt.clone().unwrap().map_err(anyhow::Error::msg)
```
fulmicoton-dd, line 208: if we unwrap, why is this an option?

chitchat/src/server.rs
```diff
@@ -144,18 +146,23 @@ pub async fn spawn_chitchat(
     let chitchat_arc = Arc::new(Mutex::new(chitchat));
     let chitchat_arc_clone = chitchat_arc.clone();
 
+    let (termination_watcher_sender, termination_watcher) = watch::channel(None);
```
fulmicoton-dd, line 149:
```suggestion
    let (termination_watcher_sender, termination_watcher) = watch::channel(None);
```

I think that codebase adopts the _tx for sender and _rx for receiver naming scheme.
It is confusing, but common in rust.

fulmicoton-dd (approved):

PR #170, opened 2025-12-05
Fixed a bug in process_message:

melsomino:
When node_states is large and the node receives a SYN with an empty digest,
    the MTU delta was incorrectly computed based on the received empty digest,
    whereas it should have used self_digest instead.
    As a result, the resulting packet size exceeded the max MTU.


fulmicoton:
Thank you!

fulmicoton (approved):

PR #172, opened 2025-12-19
Paul.masurel/only call catchup on reset

fulmicoton:


chitchat/src/lib.rs
```diff
@@ -364,6 +356,23 @@ impl Chitchat {
...
+            return;
+        }
+
+        if max_version < node_state.last_gc_version() {
+            // It is possible to have gc_version > max_version when we are catching up.
+            // If we reach this point, we probably went through a reset via chitchat gossip.
+            //
+            // Now we are trying to reset our state via grpc gossip, but the new state is already
+            // be out of date.
+            warn!(
```
fulmicoton: this is the new condition.

Before, when we gossiped with lagging nodes, we would update our state, and trigger gossip in the same round.

chitchat/src/state.rs
```diff
@@ -137,12 +138,9 @@ impl NodeState {
...
     }
 
-    // Prepare the node state to receive a delta.
-    // Returns `true` if the delta can be applied. In that case, the node state may be mutated (if a
-    // reset is required) Returns `false` if the delta cannot be applied. In that case, the node
-    // state is not modified.
+    // Assess whether the delta can be applied or not.
     #[must_use]
-    fn prepare_apply_delta(&mut self, node_delta: &NodeDelta) -> bool {
+    fn check_delta_status(&self, node_delta: &NodeDelta) -> DeltaStatus {
```
fulmicoton: I just isolated the immutable logic that considers whether we should accept a delta or not.

I also tried to make it simpler.

chitchat/src/delta.rs
```diff
@@ -38,9 +40,18 @@ impl Delta {
...
+                    // presence of an empty empty node state (due to GC).
+                    Some(DeltaOpRef::SetMaxVersion {
+                        max_version: node_delta.max_version,
+                    })
+                } else {
+                    // No need to add this operation.
+                    // It would be redundant with the last operation, max_version.
+                    None
+                }
+                .into_iter()
```
guilload: Look at you using `Option` just like a regular container :D

guilload (approved):

PR #176, opened 2026-03-02
Fix delta compatibility boundary

mikcorer:
NodeState::check_delta_status` treated `node_delta.last_gc_version == self.max_version` as incompatible because it used a strict `<` check. That equality boundary could incorrectly route a valid delta into reset handling (`ApplyAfterReset` when `from_version_excluded == 0`). During `ClusterState::apply_delta`, this could decrease `max_version` after reset and violate the monotonic invariant asserted by `monotonic_property`, causing a panic.


mikcorer:
Hi @maintainers, сould you review the fix and confirm that the approach is correct?

fulmicoton (approved):

fulmicoton:
You are correct! THank you! @mikcorer

PR #179, opened 2026-04-16
Fix monotonic property assertion failure during delta application

### LINKED ISSUE #178: Chitchat monotonicity assert failing.
This causes a panic as reported in 
https://github.com/quickwit-oss/quickwit/pull/6196

and by @rdettai

fulmicoton-dd:
## Summary
- Fixes the `monotonic_property_after >= monotonic_property_before` assertion failure in `apply_delta`
- Changes `monotonic_property()` from `(max(max_version, last_gc_version), max_version)` to `(last_gc_version, max_version)`
- Adds a regression test exercising the exact scenario that triggers the bug

## Root Cause

A race condition between concurrent gossip exchanges: between the time a node sends its digest and receives the response delta, another peer may advance its state. The responding peer computes a reset delta from the stale digest, with `last_gc_version == receiver.max_version` but `node_delta.max_version < receiver.max_version`. The old monotonic property's first element `max(mv, gc)` would remain equal while `max_version` regressed, failing the assertion.

## Fix

The new monotonic property `(last_gc_version, max_version)` is the correct invariant because `last_gc_version` only ever increases (on GC and on reset, where `check_delta_status` guarantees `node_delta.last_gc_version > self.last_gc_version`). With the first tuple element tracking `last_gc_version` independently, a reset always strictly increases it, so `max_version` can safely be lower (e.g. due to MTU truncation) without violating monotonicity.

Closes #178

### DIFF
--- a/chitchat/src/state.rs
+++ b/chitchat/src/state.rs
@@ -185,10 +185,7 @@ impl NodeState {
 
     /// Whenever we update the state, this key should always be increasing.
     pub fn monotonic_property(&self) -> (Version, Version) {
-        (
-            self.last_gc_version(),
-            self.max_version(),
-        )
+        (self.last_gc_version(), self.max_version())
     }
 
     fn reset_node(&mut self, last_gc_version: Version) {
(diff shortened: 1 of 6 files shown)


chitchat/src/state.rs
trinity-1686a, line 198: the documentation for this function isn't ideal (not related to this PR, i was just trying to understand when max_version could go down)

trinity-1686a (approved):
that's a sneaky one

PR #193, opened 2026-06-26
Compress digest

guilload:


chitchat/src/digest.rs
```diff
@@ -65,15 +108,60 @@ impl Digest {
...
+    fn serialize(&self, buf: &mut Vec<u8>) {
+        if self.compress {
+            self.serialize_v1(buf);
+        } else {
+            self.serialize_v0(buf);
+        }
+    }
+
     fn serialized_len(&self) -> usize {
+        if self.compress {
```
guilload: This is not ideal but we already follow the same "pattern" for the `Delta`.  Will fix if it starts showing up on CPU profiles.

nadav-govari (approved):

chitchat/src/digest.rs
```diff
@@ -1,8 +1,31 @@
...
+/// the legacy node count. A legacy digest never reaches this many entries (such
+/// a digest could not fit in a single datagram), so the two formats are
+/// unambiguous and the decoder can auto-detect them.
+///
+/// The magic number only exists so that V0 and V1 digests can coexist on the
+/// wire during the rollout: it tells the decoder which format it is looking at.
+/// Once every node in the cluster emits V1, there is nothing left to tell apart,
+/// so the magic number could be dropped and the first byte used directly as a
+/// format version.
+const DIGEST_MAGIC: u16 = u16::MAX;
```
fulmicoton-dd: that's not a great magic number though. Can we take any random looking number above 2^15?

guilload: yes, 65521.

guilload: prime t_t

chitchat/src/digest.rs
```diff
@@ -65,15 +108,60 @@ impl Digest {
...
+
+    /// Serializes the digest using the V1 (compressed) wire format:
+    /// `[magic: u16][format: u8][zstd block stream of (ChitchatId, NodeDigest)]`.
+    fn serialize_v1(&self, buf: &mut Vec<u8>) {
+        DIGEST_MAGIC.serialize(buf);
+        DIGEST_FORMAT_V1.serialize(buf);
+        let mut stream_writer =
+            CompressedStreamWriter::with_block_threshold(DIGEST_BLOCK_THRESHOLD);
+        for (chitchat_id, node_digest) in &self.node_digests {
+            stream_writer.append(chitchat_id);
```
fulmicoton-dd, line 147: The API between CompressedStreamWriter and deserialize_stream might be needlessly asymetric.

If we do no need the flexibility, we could have a typed CompressedStreamWriter<(ChitchatId, NodeDigest)> where we append chitchat_id and node_digest together.

chitchat/src/digest.rs
```diff
@@ -85,6 +173,16 @@ impl Serializable for Digest {
 
 impl Deserializable for Digest {
     fn deserialize(buf: &mut &[u8]) -> anyhow::Result<Self> {
+        // Peek the leading bytes without consuming them: the magic number
+        // distinguishes the V1 format from a V0 node count.
+        if buf.get(..DIGEST_MAGIC_BYTES.len()) == Some(&DIGEST_MAGIC_BYTES[..]) {
```
fulmicoton-dd: I suspect there is a starts_with no?

chitchat/src/message.rs
```diff
@@ -30,6 +30,12 @@ pub enum ChitchatMessage {
     PanicForTest,
 }
 
+impl ChitchatMessage {
+    /// Bytes written before the message body: magic number (2) + protocol
+    /// version (1) + message type (1).
+    pub(crate) const HEADER_LEN: usize = 2 + 1 + 1;
```
fulmicoton-dd: this is only true for digest today isn't it? I suspect there is a bug here.

chitchat/src/lib.rs
```diff
@@ -135,7 +135,9 @@ impl Chitchat {
                 let scheduled_for_deletion: HashSet<_> =
                     self.scheduled_for_deletion_nodes().collect();
                 let self_digest = self.compute_digest(&scheduled_for_deletion);
-                let delta_mtu = MAX_UDP_DATAGRAM_PAYLOAD_SIZE - 1 - self_digest.serialized_len();
+                let delta_mtu = MAX_UDP_DATAGRAM_PAYLOAD_SIZE
+                    - ChitchatMessage::HEADER_LEN
```
fulmicoton-dd: I don't think this is correct, the MAGIC number and version you added are included in serialized_len for instance

chitchat/src/lib.rs
```diff
@@ -153,7 +155,7 @@ impl Chitchat {
                     self.scheduled_for_deletion_nodes().collect::<HashSet<_>>();
                 let delta = self.cluster_state.compute_partial_delta_respecting_mtu(
                     &digest,
-                    MAX_UDP_DATAGRAM_PAYLOAD_SIZE - 1,
+                    MAX_UDP_DATAGRAM_PAYLOAD_SIZE - ChitchatMessage::HEADER_LEN,
```
fulmicoton-dd: no there a no Syn in the message and HEADER_LEN include the syn magic/version.

fulmicoton-dd:
ChitchatMessage has versioning. I suggest we rely on that instead of having versioning and magic number for each of its component

chitchat/src/digest.rs
guilload: Good catch. Why didn't I think of that t_t

Update: ended up using `strip_prefix`.

chitchat/src/message.rs
guilload: I think what happened is that I added a protocol version in `d9049ab` and you added a magic in `f783620` and we forgot to update the few places where we compute the space left in the datagram

chitchat/src/message.rs
```diff
@@ -17,10 +17,18 @@ const MAGIC_NUMBER: u16 = 45_139;
...
+    Syn {
+        protocol_version: ProtocolVersion,
+        cluster_id: String,
+        digest: Digest,
+    },
     /// Scuttlebutt SYN-ACK: node B returns a partial update as described in the Scuttlebutt
     /// reconciliation algorithm and its own digest.
-    SynAck { digest: Digest, delta: Delta },
+    SynAck {
+        protocol_version: ProtocolVersion,
```
fulmicoton-dd: I think it might be better to separate the version from the model.

```rust
struct VersionedChitchatMessage {
    version: ProtocolVersion,
    message: ChitchatMessage,
}
```

(I said enum in the call, that was dumb).

fulmicoton-dd: We might want to remove the serialize bullshit to a separate module if it is getting larger with the different version too.

fulmicoton-dd (approved):
LGTM but I think it would be worth doing putting version outside of the model

chitchat/src/message.rs
```diff
@@ -30,238 +28,80 @@ pub enum ChitchatMessage {
...
 
-impl ProtocolVersion {
-    pub fn from_code(code: u8) -> Option<Self> {
-        match code {
-            0 => Some(Self::V0),
-            _ => None,
-        }
+impl ChitchatFrame {
+    pub(crate) fn new(version: ProtocolVersion, message: ChitchatMessage) -> Self {
+        Self { version, message }
```
fulmicoton-dd: nitpick:
version and message are public. I would have remove new and built the struct directly maybe.

chitchat/src/message.rs
```diff
@@ -30,238 +28,80 @@ pub enum ChitchatMessage {
...
-#[derive(Clone, Copy, Eq, PartialEq)]
-#[repr(u8)]
-enum ProtocolVersion {
-    V0 = 0,
+/// A [`ChitchatMessage`] paired with the protocol version it is encoded
+/// with. The version is not part of the message model (see
+/// [`ChitchatMessage`]'s doc); this struct is the unit that actually travels
+/// on the wire, see `message_serialize`.
+#[derive(Debug, Eq, PartialEq)]
+pub struct ChitchatFrame {
```
fulmicoton-dd: nitpick:
VersionedChitchatMessage ?

I usually see "frame" in network protocol as the unit of stuff you transfer a stream of for something higher level.

fulmicoton-dd (approved):

PR #195, opened 2026-07-01
Bugfix on grpc chitchat reset causing assert failure

### LINKED ISSUE #194: assertion failed: monotonic_property_after > monotonic_property_before triggered
On chitchat version 0.11.0 in Cloudprem v0.1.30 (aarch64-unknown-linux-gnu 2026-06-29T10:47:20Z 6c19cc9)

fulmicoton:
Chitchat offers an API to reset a node state.
Quickwit uses it to bring node to an up to date
state using a gRPC call.

We observed an assert failure of the monotonic assert in prod.

There are in fact two possible cause of failure:
- a race condition between a chitchat reset and the gRPC. In that case, the grpc reset would bring us to an anterior gc version.

- a node max version that is higher that the highest KV version. This is possible if the KV with max version was GC'ed. In that case, we were forgetting to set max version and ended up with breaking the monotonic contraint.

Both failure mode are now caught in a unit test.

The solution is not trying to be smart.
We now explicitly reject grpc calls that would not strictly increase the monotonic property.

We kept the rejection on gRPC reset that do not increase max_version too.

We also do set max_version at the end of the call.

Closes #194

### DIFF
--- a/chitchat/src/lib.rs
+++ b/chitchat/src/lib.rs
@@ -35,6 +35,22 @@ pub use crate::server::{ChitchatHandle, spawn_chitchat};
 use crate::state::ClusterState;
 pub use crate::types::{ChitchatId, DeletionStatus, Heartbeat, Version, VersionedValue};
 
+/// Sets up a tracing subscriber writing to the test harness's captured
+/// output, so `cargo test` can display logs on failure.
+///
+/// Guarded by a `OnceLock` because `tracing::subscriber::set_global_default`
+/// can only be called once per process, and all tests in a given test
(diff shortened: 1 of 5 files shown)


chitchat/src/lib.rs
```diff
@@ -355,29 +355,25 @@ impl Chitchat {
             return;
         };
 
-        if node_state.max_version() >= max_version {
```
fulmicoton, line 358: this was technically not a bad idea but not really necessary either so i removed it

trinity-1686a (approved):

